PR #8, opened 2017-07-14
Use bisect for weather temperature ranges and reset text color properly

jvperrin:
The black text previously gets hidden when on a dark background, so reset to whatever the default text color is instead.

I'm conflicted on whether this is actually more readable than it was before since `bisect` is more complicated than just a bunch of `if`/`else` blocks. I do think it's reasonably clear though once the documentation about it has been read. Thoughts?

### DIFF
--- a/ircbot/weather.py
+++ b/ircbot/weather.py
@@ -1,4 +1,5 @@
 import urllib.parse
+from bisect import bisect
 
 import requests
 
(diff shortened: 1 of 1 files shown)


jvperrin:
I'm working on changing this to use a dictionary with a tip from @asottile, should make it a bit cleaner.

kpengboy:
It will likely be much slower...

nattofriends:
Oh no, the ultra performant weather function will now be a bottleneck. Whatever shall we do?

kpengboy:
Actually, after @asottile's tip I'd have to say it's more readable

ircbot/weather.py
```diff
@@ -17,24 +18,29 @@ def icon(temp):
...
+    # also means the first value matches anything under it.
+    temp_ranges = {
+        40: '\x0312',  # Light blue (< 40)
+        50: '\x0311',  # Light cyan
+        60: '\x0310',  # Teal
+        70: '\x0314',  # Grey
+        75: '\x0F',   # Reset (default text color)
+        80: '\x0307',  # Orange
+        90: '\x0305',  # Maroon
+        999: '\x0304',  # Red (> 90)
```
kpengboy: Well, one could use `float('inf')` or `1e2000` as well...

asottile: if the weather is 999F, we have bigger problems than an irc bot

kpengboy: Global warming is a serious problem, and we would be doing our part to raise awareness :P

ircbot/weather.py
```diff
@@ -17,24 +18,29 @@ def icon(temp):
...
+    }
+    temps = sorted(temp_ranges)
```
kpengboy: Looking forward to Python 3.6, when we can get rid of this line...

kpengboy: Looking forward to Python 3.6, when we can avoid a sort operation...

asottile: only an implementation detail in py36 (though unlikely to change) -- I wouldn't start depending on the behaviour until it's standards track for py37

jvperrin:
Wow, I didn't expect this much feedback, thanks 😂

PR #12, opened 2017-07-23
Add quotes

chriskuehl:
Still need to backfill the old ones into the DB, but the code works.


chriskuehl:
I imported all of the old quotes that I have from IRC logs, looks like it's most of them. Hopefully I can grab the rest from waf.

PR #17, opened 2017-09-14
get inspired

tahabi:
let create inspire you with a fine selection of quotations from such luminaries as DJ Khaled, NPCs from "Sid Meier's Alpha Centauri," the top three Google search results for "inspirational quotes.", and [inspirobot.me](https://inspirobot.me)

in addition, create can now help answer the eternal question: is jvperrin in the lab?

### DIFF
--- a/ircbot/inspire.py
+++ b/ircbot/inspire.py
@@ -1,0 +1,23 @@
+from ircbot import db
+
+
+def rand(mysql_password):
+    with db.cursor(password=mysql_password) as c:
+        c.execute(
+            'SELECT text FROM inspire ORDER BY RAND() LIMIT 1',
+        )
+
+        return c.fetchone()['text']
+
+
+def like(mysql_password, term):
+    with db.cursor(password=mysql_password) as c:
+        c.execute(
+            'SELECT text FROM inspire WHERE LOWER(text) LIKE LOWER(%s) ORDER BY RAND() LIMIT 1',
+            '%{}%'.format(term),
+        )
+        result = c.fetchone()
+        if result is None:
+            return "No quotes matching '{}' found.".format(term)
+        else:
+            return result['text']
--- a/ircbot/inspire.py
+++ b/ircbot/inspire.py
@@ -1,23 +1,16 @@
 from ircbot import db
 
 
-def rand(mysql_password):
+def inspire(mysql_password, msg):
+    search = msg.strip().split(' ')
+    term = ' '.join(search[1:]) if len(search) > 1 else ''
+
     with db.cursor(password=mysql_password) as c:
         c.execute(
-            'SELECT text FROM inspire ORDER BY RAND() LIMIT 1',
+            'SELECT text FROM `inspire` ' +
+            'WHERE LOWER(text) LIKE LOWER(%s) ' +
+            'ORDER BY RAND() LIMIT 1',
+            '%{}%'.format(term),
         )
 
         return c.fetchone()['text']
-
-
-def like(mysql_password, term):
-    with db.cursor(password=mysql_password) as c:
-        c.execute(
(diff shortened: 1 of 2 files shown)


ircbot/ircbot.py
```diff
@@ -23,14 +23,15 @@
...
+from ocflib.lab.stats import staff_in_lab
 
 from ircbot import debian_security
 from ircbot import emoji
+from ircbot import inspire
 from ircbot import quotes
 from ircbot import rackspace_monitoring
 from ircbot import weather
 
-
```
matthew-mcallister: ?

ircbot/ircbot.py
```diff
@@ -152,10 +157,12 @@ def new(_):
                             break
                     except re.error:
                         continue
+
```
matthew-mcallister: These newlines pass the style test so I don't mind, but you missed the `elif` above this one.

matthew-mcallister:
We should probably eventually decide when these new "features" are getting to be too much. I do think the `staff_in_lab` one is genuinely useful, however.

ircbot/inspire.py
```diff
@@ -0,0 +1,23 @@
...
+    with db.cursor(password=mysql_password) as c:
+        c.execute(
+            'SELECT text FROM inspire WHERE LOWER(text) LIKE LOWER(%s) ORDER BY RAND() LIMIT 1',
+            '%{}%'.format(term),
+        )
+        result = c.fetchone()
+        if result is None:
+            return "No quotes matching '{}' found.".format(term)
+        else:
+            return result['text']
```
jvperrin: These two functions are pretty similar, could they be combined into one function or something? They both execute a DB query and then use the first result and the `text` column.

ircbot/ircbot.py
```diff
@@ -131,9 +134,11 @@ def respond(msg, ping=True):
             elif shrug:
                 width = len(shrug.group(1))
                 respond('¯\\' + ('_' * width) + '(ツ)' + ('_' * width) + '/¯', ping=False)
+
             elif msg.startswith((IRC_NICKNAME + ' ', IRC_NICKNAME + ': ')):
                 command, *args = msg[len(IRC_NICKNAME) + 1:].strip().split(' ')
                 self.handle_command(is_oper, command.lower(), args, respond)
+
```
chriskuehl: ? what's with all the random newlines?

ircbot/ircbot.py
```diff
@@ -166,6 +173,23 @@ def new(_):
...
+                        break
+                else:
+                    respond('{} is not in the lab'.format(username))
+
+            elif msg.startswith('!inspire'):
+                search = msg.strip().split(' ')
+                if len(search) > 1:
+                    respond(inspire.like(self.mysql_password, ' '.join(search[1:])))
+                else:
+                    respond(inspire.rand(self.mysql_password))
```
jvperrin: Changing the functions in the part above should make this a bit simpler too

ircbot/ircbot.py
jvperrin: I think it's to try to separate out the functions

chriskuehl: the `elif`s don't do that already? lol

ircbot/ircbot.py
tahabi: oops thanks

ircbot/ircbot.py
tahabi: just makes it a bit more readable I feel

chriskuehl: I think the "real" solution is to improve the way we dispatch out commands to other functions/modules rather than having a big if-elif tree. I don't really oppose the random blank lines, but it does look quite bizarre to me. I think it's probably a smell that the code isn't organized super well right now.

tahabi: Agreed. Sounds like a fun project for someone's project team.

PR #19, opened 2017-09-17
Refactor commands out into independent plugins

### LINKED ISSUE #18: Split different bot responses out into functions
As suggested as part of the feedback for #17, splitting up the different ircbot responses into separate functions would be a good thing to do to refactor how the bot works.

chriskuehl:
Fixes #18 

Each set of commands is now a "plugin". They're plugins because the core bot code discovers and allows them to register hooks (currently just listeners) and otherwise has no logic for them.

Eventually we can expand this pattern to also register timers so that we can move that code out of ircbot.py too. Ideally we'd also move the Celery listeners for create into its plugin file.

A simple plugin looks like this:

```python
"""Show your appreciation."""
import random


def register(bot):
    bot.listen(
        r'^thanks', thanks, require_mention=True,
        help='thank create for being helpful',
    )
    bot.listen(
        r'^thank (.*)$', thank_someone, require_mention=True,
        help='have create thank somebody on your behalf',
    )


def thanks(text, match, bot, respond):
    respond(random.choice((
        "you're welcome",
        'you are most welcome',
        'any time',
        'sure thing boss',
    )))


def thank_someone(text, match, bot, respond):
    respond('thanks, {}!'.format(match.group(1)), ping=False)
```

The docstrings and help text are not currently used, but it will be easy to turn those into automatically-generated documentation in the future.


ircbot/plugin/magic.py
```diff
@@ -0,0 +1,24 @@
...
+def _magic(thing):
+    return '(ノﾟοﾟ)ノﾐ★゜・。。・゜゜・。{} 。・゜☆゜・。。・゜'.format(thing)
+
+
+def magic(text, match, bot, respond):
+    respond(_magic(match.group(1) or 'magic'), ping=False)
+
+
+def mystery(text, match, bot, respond):
+    respond(_magic('https://mystery.fuqu.jp/'), ping=False)
```
matthew-mcallister: Nice.

chriskuehl:
1. I was worried about if we ever re-use functions or construct functions on-demand or something. E.g. as a contrived example, say we had a plugin that responds with an emoji for "debian" and another emoji for "penguin", we might write something like:
   ```python
   def register(bot):
       bot.listen(r'debian', _respond_fn('🍥'), help='print a debian emoji')
       bot.listen(r'penguin', _respond_fn('🐧'), help='print a penguin emoji')

   def _respond_fn(emoji):
       def wrapped(text, match, bot, respond):
           respond(emoji)
       return wrapped
   ```
   It would be kinda annoying (but not impossible) to shove docstrings onto this. Open to suggestions.
2. I agree, the set of params it passes are kinda arbitrary and it's annoying to add more stuff (like the person who actually triggered it). I was thinking of making a namedtuple like `MatchedMessage` with params like `text`, `match`, `nickname`, `is_oper` and a function `respond`. Then we'd just pass in `bot` and `msg` as parameters to all the plugin functions. I think this may be a bit cleaner and also easier to expand in the future if we find we need more things in the tuple.

chriskuehl:
On second thought for 1, I think docstrings are pretty reasonable. I'll change it to do that.

matthew-mcallister:
1. That's a fair point, but you could use wrapper functions to achieve this functionality, including having different text for the in-IRC help and the actual code documentation.
2. That sounds like a great idea, much more extensible. I say go ahead with that.

chriskuehl:
@matthew-mcallister now updated with both of those changes. still not actually exposing the help anywhere, but the data is there.

ircbot/ircbot.py
```diff
@@ -108,155 +161,42 @@ def on_pubmsg(self, conn, event):
...
+                self.say(event.target, full_raw_text)
+
+            was_mentioned = raw_text.startswith((IRC_NICKNAME + ' ', IRC_NICKNAME + ': '))
+
+            for listener in self.listeners:
+                text = raw_text
+                if listener.require_mention:
+                    if was_mentioned:
+                        # Chop off the bot nickname.
+                        text = ' '.join(text.split(' ')[1:])
```
matthew-mcallister: Wouldn't it be more accurate to do
```
text.split(' ', 1)[1]
```
Or, if you wanted to chop off _all_ the leading whitespace,
```
re.split('\s+', text, 1)[1]
```

chriskuehl: yeah, any of these are probably fine. i'll pick the first

PR #22, opened 2017-12-21
make ping case insensitive

kkuehler:


ircbot/plugin/ping.py
```diff
@@ -3,12 +3,16 @@
...
 
 def pong(bot, msg):
     pleased = random.randint(0, 99)
     if pleased:
-        msg.respond(msg.text.replace('i', 'o'), ping=False)
+        msg.respond(
+            ''.join([chr(ord(c) + 6)
+                     if c.lower() == 'i' else c
+                     for c in msg.text]),
+            ping=False)
```
jvperrin: I'd do something like `msg.respond(msg.text.translate(str.maketrans('iI', 'oO')), ping=False)`. Seems a lot easier anyway for this kind of thing.

ircbot/plugin/ping.py
```diff
@@ -3,12 +3,16 @@
 
 
 def register(bot):
-    bot.listen(r'^p+i+n+g+$', pong, require_mention=True)
+    bot.listen(r'^[Pp]+[Ii]+[Nn]+[Gg]+$', pong, require_mention=True)
```
jvperrin: Instead of doing this, I'd add an option to [`bot.listen` that gets passed to `re.compile`](https://github.com/ocf/ircbot/blob/9c4592501a5f82de32982b72fe7b61156cc505cd/ircbot/ircbot.py#L137) to make it accept regex flags, like `re.IGNORECASE`, and then that'll fix this and be useful for future regexes.

ircbot/plugin/ping.py
chriskuehl: or maybe:

```python
msg.text.replace('i', 'o').replace('I', 'O')
```

the benefit here being that it's obvious what's going on :)

i'm a big fan of "dumb" solutions like this even when there's a smarter way to do things. in this case, it's not important to be performant or general or anything.

jvperrin: Yeah, I considered that one too, I should have mentioned it. It's definitely pretty clear what is going on, which is nice.

kkuehler:
Should be good, @jvperrin, @chriskuehl

ircbot/plugin/ping.py
```diff
@@ -1,14 +1,17 @@
...
+    bot.listen(r'^p+i+n+g+$', pong, flags=re.IGNORECASE, require_mention=True)
 
 
 def pong(bot, msg):
     pleased = random.randint(0, 99)
     if pleased:
-        msg.respond(msg.text.replace('i', 'o'), ping=False)
+        msg.respond(
+            msg.text.replace('i', 'o').replace('I', 'O'),
+            ping=False)
```
jvperrin: I'd just put this parenthesis on the next line, then you can have a trailing comma after `ping=False`, which is a bit nicer if you want to add additional arguments later:

    msg.respond(
        msg.text.replace('i', 'o').replace('I', 'O'),
        ping=False,
    )

kkuehler: done!

PR #27, opened 2018-01-13
Add a help page

### LINKED ISSUE #1: support help command
e.g. "create: help"

chriskuehl:
Fixes #1

Running at http://supernova.ocf.berkeley.edu:8888/ (will leave this up for a while)

The page is pretty rough right now, but I'm all about shipping something which is at least minimally useful and improving it later :)

The plan is to have the ircbot itself serve this page (the code in this branch does that) and add configs to marathon/lb nginx to route ircbot.o.b.e there.

### DIFF
--- a/ircbot/plugin/help.py
+++ b/ircbot/plugin/help.py
@@ -1,0 +1,51 @@
+"""Provide help information."""
+import collections
+import http.server
+import threading
+
+import jinja2
+
+
+def register(bot):
+    threading.Thread(target=help_server, args=(bot,)).start()
+    bot.listen(r'^help$', help, require_mention=True)
+
+
+def help(msg, bot):
+    """Provide a link to this help page."""
+    msg.respond('https://ircbot.ocf.berkeley.edu/')
+
+
+def build_request_handler(bot):
+    jinja_env = jinja2.Environment(
+        loader=jinja2.PackageLoader('ircbot', ''),
+        autoescape=True,
+    )
+
+    class RequestHandler(http.server.BaseHTTPRequestHandler):
+
+        def do_GET(self):
+            if self.path != '/':
+                self.send_response(404, 'File not found')
+                self.end_headers()
+                self.wfile.write(b'404 File not found')
+            else:
+                plugins = collections.defaultdict(set)
+                for listener in bot.listeners:
+                    plugins[bot.plugins[listener.fn.__module__]].add(listener)
+                rendered = jinja_env.get_template('plugin/templates/help.html').render(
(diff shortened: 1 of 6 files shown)


ircbot/plugin/help.py
```diff
@@ -0,0 +1,51 @@
...
+                self.wfile.write(b'404 File not found')
+            else:
+                plugins = collections.defaultdict(set)
+                for listener in bot.listeners:
+                    plugins[bot.plugins[listener.fn.__module__]].add(listener)
+                rendered = jinja_env.get_template('plugin/templates/help.html').render(
+                    plugins=sorted(plugins.items(), key=lambda p: p[0].__name__),
+                ).encode('utf8')
+
+                self.send_response(200, 'Okay')
```
kpengboy: I would use the standard response message...

ircbot/plugin/help.py
```diff
@@ -0,0 +1,51 @@
...
+                self.send_header('Content-Length', len(rendered))
+                self.send_header('Content-Type', 'text/html; charset=utf-8')
+                self.end_headers()
+                self.wfile.write(rendered)
+
```
kpengboy: Might want to consider supporting the HEAD method as well? I think many clients expect that method to exist

ircbot/plugin/templates/help.html
```diff
@@ -0,0 +1,38 @@
...
+    <head>
+        <title>ocf ircbot help</title>
+        <style>
+            tt { background-color: #eee; padding: 3px; }
+        </style>
+    </head>
+    <body>
+        <h1>create: <em>an IRC bot for doing stupid stuff and sometimes handling commands for account creation</em></h1>
+        <hr />
+
```
kpengboy: I would add a line like "Available commands are:" here for context

ircbot/plugin/help.py
chriskuehl: why?

ircbot/plugin/help.py
chriskuehl: HEAD is probably a nice-to-have. I've done a bunch of smaller http servers in this pattern before and no clients have had an issue with it. Feel free to add it though.

ircbot/plugin/templates/help.html
chriskuehl: added. though i suspect we might change all of this later anyway if we put effort into making the page look better.

chriskuehl:
Another problem is that, as-written, we won't be able to run two dev ircbots on one host because they'll both want to bind to this port. Not sure what the best way to handle that is yet.

PR #28, opened 2018-01-25
Log calnet UID and callink OID, move logging to plugin

jvperrin:
This is for [rt#6971](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=6971). It logs lines like `jvperrin created (Jason Perrin, Calnet UID: 1101587)` in #atool. I also moved all the logging stuff to the `create` plugin, since there was a comment about doing that.

### DIFF
--- a/ircbot/plugin/create.py
+++ b/ircbot/plugin/create.py
@@ -1,7 +1,13 @@
 """Approve accounts."""
 import random
+import ssl
 
 from celery import exceptions
+from celery.events import EventReceiver
+from kombu import Connection
+
+from ircbot.ircbot import IRC_CHANNELS_ANNOUNCE
+from ircbot.ircbot import IRC_CHANNELS_OPER
 
 
 def register(bot):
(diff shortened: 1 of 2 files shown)


ircbot/plugin/create.py
```diff
@@ -50,3 +56,77 @@ def flip(bot, msg):
...
+    def bot_announce(targets, message):
+        for target in targets:
+            bot.say(target, message)
+
+    def on_account_created(event):
+        request = event['request']
+
+        if request['calnet_uid']:
+            calnet_id = 'Calnet UID: {}'.format(request['calnet_uid'])
+        elif request['calnet_oid']:
```
chriskuehl: are you sure this shouldn't be callink oid?

jvperrin: Hehe nope, good catch, I didn't realize they were named differently like that. I don't think any group accounts have been made recently while I was doing this for this part to be tested unfortunately.

jvperrin: Ok, this should be fixed now.

PR #31, opened 2018-02-07
create wants you to ship your code

kkuehler:
could be too spammy but i find it amusing

### DIFF
--- a/ircbot/plugin/shipit.py
+++ b/ircbot/plugin/shipit.py
@@ -1,0 +1,18 @@
+"""Your code is probably not ready for production."""
+import re
+
+
+def register(bot):
+    bot.listen(r's+h+i+p+\s*i+t+', shipit, flags=re.IGNORECASE)
+
+
+def shipit(bot, msg):
+    """shipit anyway!"""
+    msg.respond('.  o ..', ping=False)
+    msg.respond('    o . o o.o', ping=False)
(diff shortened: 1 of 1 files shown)


jvperrin:
(source is at https://github.com/chriskuehl/shipit/blob/master/index.html)

ircbot/plugin/shipit.py
```diff
@@ -0,0 +1,18 @@
+"""Your code is probably not ready for production."""
+import re
+
+
+def register(bot):
+    bot.listen(r's+h+i+p+i+t+', shipit, flags=re.IGNORECASE)
```
chriskuehl: maybe allow an optional space between the "ship" and "it"? e.g. `'s+h+i+p+\s*i+t+'`

kkuehler:
Sad days. Unfortunately the beautiful ascii art does not mirror to Slack well. Since the majority of OCFers use Slack it does not make much sense to have this feature as it would lead to confusion. First, Slack truncates leading whitespace; I was able to get around this by adding the zero-width space character to the beginning of every line but Slack still had weird formatting issues and it didn't line up. You can put it in code tags for a passable look but it mirrors slowly and looks pretty bad. Plus, an IRC bot should not have code tags to accommodate another platform.

dkess:
sad reax only

nattofriends:
Have the bridge add the code tags...!

PR #34, opened 2018-04-18
magic: explain solutions to technical problems

kpengboy:


ircbot/plugin/magic.py
```diff
@@ -3,7 +3,7 @@
 
 def register(bot):
     bot.listen(r'^magic ?(.*)$', magic, require_mention=True)
-    bot.listen(r'\bmystery\b', mystery)
+    bot.listen(r"\bmystery\b|^why (doesn't it work|isn't it working)$|^how does it work$", mystery)
```
chriskuehl: maybe also add "how does anything (even)? work"

PR #36, opened 2018-04-19
theory_practice: everything is broken

kpengboy:


ircbot/plugin/theory_practice.py
```diff
@@ -0,0 +1,19 @@
...
+
+def register(bot):
+    bot.listen(r"^why (doesn't anything work|isn't anything working)$",
+               theory_practice, require_mention=True)
+
+
+def theory_practice(bot, msg):
+    # Slightly modified from a sign written by daradib or sanjayk or whoever
+    # As of this writing, it's still posted in the OCF lab
+    # Original can be found at https://www.ocf.berkeley.edu/~daradib/dl/ocf/publicity/signs/theory-practice.odt
```
chriskuehl: this function should have a docstring, since this is what we use to populate the help text on https://ircbot.ocf.berkeley.edu/

PR #38, opened 2018-05-20
fix small bug where not every DSA line ends in 'security update'

abizer:


ircbot/plugin/debian_security.py
```diff
@@ -19,10 +19,12 @@ def dsa_list():
     for item in root.iter('{http://purl.org/rss/1.0/}item'):
         # title is of the form "DSA-3804 linux - security update"
         title = item.find('{http://purl.org/rss/1.0/}title').text
-        m = re.match('DSA-(\d+) (.+?) - security update$', title)
+
+        # group 1: dsa num, group 2: package or None, group 3: inline description
+        m = re.match('DSA-(\d+)(.+)? - (.+?)$', title)
```
kkuehler: I imagine you still want that space between the DSA and the package, or your package will always have leading whitespace. You can probably get rid of `m.group(2).strip()` below then too. Also, why match for "security update" now that it's inconsistent -- we don't actually care if it's there or not. It was probably just there to anchor the regex.

nit: Also, for the package regex, I don't think it hurts to be [more specific](https://github.com/jtniehof/dpkg/blob/03b4300e51cec35e49e38aef14d72acb6adea6bb/lib/dpkg/parsehelp.c#L130-L148)

abizer: well, we don't actually need the space between the DSA and package name, we can just mark it optional in the regex and the match will fall through to the space in ' - ' if necessary. This also takes care of the leading whitespace.

PR #39, opened 2018-05-29
Fix api break in github3

asottile:
https://github.com/sigmavirus24/github3.py/commit/0244423a5e6a8


asottile:
since https://github.com/ocf/ircbot/commit/5c9510fe17285ae94be104774016f416490547d0 this was incorrect it seems :(

PR #41, opened 2018-06-14
Revert "Make trailing delimiter on replacements optional"

asottile:
This reverts commit fc5e2da06f972ace7fb2ff2335aac06cd106314e.

@jvperrin @chriskuehl this seems to fix this for me.  I'm really not sure why though

### DIFF
--- a/ircbot/plugin/regex.py
+++ b/ircbot/plugin/regex.py
@@ -3,7 +3,7 @@ import re
 
 
 def register(bot):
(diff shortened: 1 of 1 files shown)


chriskuehl:
Seems like we may need to follow-up to make omitting trailing delimiters work again, but this is better than the status quo where having them is broken :)

jvperrin:
Yeah, agreed this is better than having them broken but I'm not sure why it wasn't working and need to investigate more when I get back to an actual computer (in ~ 10 days).

PR #45, opened 2018-07-21
regex: honor 'g' flag

kpengboy:
This changes the behavior of regex to be more in line with programming languages which have the same function. Namely, `s/foo/bar` and `s/foo/bar/` now substitute only the first match; the global flag `g` makes it substitute all matches.

### DIFF
--- a/ircbot/plugin/regex.py
+++ b/ircbot/plugin/regex.py
@@ -3,12 +3,13 @@ import re
 
 
 def register(bot):
-    bot.listen(r'(?:^| )s([!@"#$%&\'*./:;=?\\^_`|~])(.+?)\1(.*?)\1?g?$', replace)
+    bot.listen(r'(?:^| )s([!@"#$%&\'*./:;=?\\^_`|~])(.+?)\1(.*?)(?:$|\1(g?))$', replace)
 
 
 def replace(bot, msg):
     """Regex-replace some text."""
     old = msg.match.group(2)
+    is_global = (msg.match.group(4) == 'g')
 
     # By default, re.sub processes strings like r'\n' for escapes,
     # turning it into an actual newline. By using a function for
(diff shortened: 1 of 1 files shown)


ircbot/plugin/regex.py
```diff
@@ -3,12 +3,13 @@
...
 
 def register(bot):
-    bot.listen(r'(?:^| )s([!@"#$%&\'*./:;=?\\^_`|~])(.+?)\1(.*?)\1?g?$', replace)
+    bot.listen(r'(?:^| )s([!@"#$%&\'*./:;=?\\^_`|~])(.+?)\1(.*?)(?:$|\1(g?))$', replace)
 
 
 def replace(bot, msg):
     """Regex-replace some text."""
     old = msg.match.group(2)
+    is_global = (msg.match.group(4) == 'g')
```
kkuehler: Well, if you are gonna use parens, might as well go the full course and lose the if-statement below.

```python
gflag = int(msg.match.group(4) != 'g')
```

nattofriends:
Although it's nice to have this behavior align with regexes in code, it is hard for me to imagine a human wanting to only replace the first instance of a prior line of text.

ethanhs:
@nattofriends here is an common one: misspelling an article

s/ an / a /

Though I do suppose one could always give more context. I'd personally prefer to have it align with other languages, but I wont lose sleep over it.

kpengboy:
@chriskuehl @asottile What are your votes on this proposal?

chriskuehl:
shipit, sure. not worth bikeshedding about this.

jvperrin:
Yeah if most people are for it, then it sounds like they expect this behavior, which contradicts the assumption I had that people would expect a global replacement with or without the `g` flag. I'm fine with this being shipped too, I just don't see it being very useful overall, but go for it.

kpengboy:
...Since there aren't strong opinions on this proposal, and it is a breaking change, I'm shelving it for now, pending further observation on usage patterns.

jvperrin:
I mean I think shipping this is fine, it's only an IRC regex replacement change, but closing it for now is ok too

PR #49, opened 2018-07-26
Add !g command for Googling

chriskuehl:


ircbot/plugin/google.py
```diff
@@ -0,0 +1,32 @@
...
+
+def google(bot, msg):
+    """Search Google."""
+    query = msg.match.group(1)
+
+    resp = requests.get(
+        'https://www.googleapis.com/customsearch/v1',
+        params={
+            'key': bot.googlesearch_key,
+            'cx': bot.googlesearch_cx,
```
jvperrin: What does `cx` stand for here? It's not clear to me what this credential actually represents, I think adding a comment about that it is might help a bit.

ircbot/plugin/google.py
```diff
@@ -0,0 +1,32 @@
...
+            'q': query,
+        }
+    )
+    if resp.status_code != 200:
+        msg.respond('error searching google: {.status_code}'.format(resp), ping=False)
+    else:
+        results = resp.json()
+        if len(results.get('items', ())) == 0:
```
jvperrin: The `items` key doesn't even exist in the results if there are no results?

ircbot/plugin/google.py
```diff
@@ -0,0 +1,32 @@
...
+            msg.respond('no results :(', ping=False)
+        else:
+            item = results['items'][0]
+            msg.respond(
+                '\x0314{0[title]} \x0F| \x0302{0[link]}'.format(item),
```
jvperrin: I feel like text colors should be (eventually) factored out into a separate function, maybe in a utils file or something, since they are already used for the [weather](https://github.com/ocf/ircbot/blob/ef8c2525530a7b7c59db76f42c684b0602c40fd5/ircbot/plugin/weather.py#L36), [stack exchange](https://github.com/ocf/ircbot/blob/5ca842996824c2ec92d2925d373d0a4e8e6dce28/ircbot/plugin/stack_exchange.py#L93-L95), and [rackspace](https://github.com/ocf/ircbot/blob/b6abdcbabf8bf1660989416cd6d03ab050c2e396/ircbot/plugin/rackspace_monitoring.py#L95) plugins, and it's not clear each time what colors they represent. Anyway, more like future work to do than anything with this PR in particular.

ircbot/plugin/google.py
chriskuehl: it's the "id" of the custom search engine, i'll add a comment

ircbot/plugin/google.py
chriskuehl: correct :\

ircbot/plugin/google.py
chriskuehl: agree

ircbot/plugin/google.py
jvperrin: Sad :'(

Well, this is a good workaround for it at least :+1:

ircbot/plugin/google.py
jvperrin: Ah, that makes more sense, thanks for adding a comment on this.

ircbot/plugin/google.py
kpengboy: I would recommend `if not results.get('items')` here.

chriskuehl: I know, we've discussed this before. I think the readability of `len(x) == 0` is better than `not x`, as it makes clear that we are dealing with sequences.

chriskuehl: Thanks, I'm aware, we've had this discussion several times already.

ethanhs: >PEP 8 is not a bible, and even explicitly says to ignore guidelines when following them would decrease readability.

Certainly, but "decreases readability" sounds a lot like opinion to me :)

I honestly don't care enough to argue over this however. As long as we are consistent.

chriskuehl: Yes, it's my opinion :). If you want to change the code, then go ahead, but my view is that it makes more sense to ask "is the length of the list zero" than "is not the list", and I think most people would agree with that.

PR #54, opened 2018-09-04
Phabulous Macros @ OCF

abizer:
this brings a simple adaptation of phabricator macros\[0\] to OCF. they are a staple at facebook

now you can do something like
macroexample

to add do `#m add <slug> <link>` and delete `#m delete <slug>`

\[0\]: https://secure.phabricator.com/project/profile/283/

### DIFF
--- a/ircbot/plugin/macros.py
+++ b/ircbot/plugin/macros.py
@@ -1,0 +1,64 @@
+"""Provide OCF image macros, inspired by Phabricator Macros"""
+from ircbot import db
+
+
+def register(bot):
+    bot.listen(r'^#m (\w+)$', show)
+    bot.listen(r'#m add (\w+) (.+)$', add)
+    bot.listen(r'#m delete (\w+)$', delete)
+
+
+def show(bot, msg):
+    """Return a macro by slug."""
+    slug = msg.match.group(1)
+
+    with db.cursor(password=bot.mysql_password) as c:
+        c.execute(
+            'SELECT link FROM macros WHERE slug = %s',
+            (slug,)
+        )
+        macro = c.fetchone()
+        if macro is not None:
+            msg.respond(macro['link'], ping=False)
+        else:
+            msg.respond('macro `{}` does not exist.'.format(slug))
+
+
+def add(bot, msg):
+    """Add a new macro."""
+
+    slug = msg.match.group(1)
+    link = msg.match.group(2)
+
+    if len(link) > 80:
+        msg.respond('please try to keep macro links below 80 characters')
+
+    with db.cursor(password=bot.mysql_password) as c:
+
+        c.execute('SELECT * FROM macros WHERE slug = %s', (slug,))
+        result = c.fetchone()
+        if result is not None:
+            msg.respond(
+                'macro `{}` already exists as {}'.format(
+                    result['slug'],
+                    result['link'],
+                )
+            )
+        else:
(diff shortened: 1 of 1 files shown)


ircbot/plugin/macros.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def add(bot, msg):
+    """Add a new macro."""
+
+    slug = msg.match.group(1)
+    link = msg.match.group(2)
+
+    if len(slug) > 50 or len(link) > 100:
+        msg.respond('macro slugs must be < 50 and links < 100 characters')
```
chriskuehl: 100 chars seems pretty low for a link tbh

abizer: it's intentionally small to keep links from breaking across a line on IRC, although I'm amendable to changing the specific number

ircbot/plugin/macros.py
```diff
@@ -0,0 +1,68 @@
...
+            msg.respond('macro added as `{}`'.format(slug))
+
+
+def delete(bot, msg):
+    """Delete a macro."""
+
+    slug = msg.match.group(1)
+    with db.cursor(password=bot.mysql_password) as c:
+        c.execute(
+            'DELETE FROM macros WHERE slug = %s',
```
kpengboy: Do we care about malicious users deleting slugs?

ircbot/plugin/macros.py
kpengboy: Also it should be `<=`

kpengboy:
Now we just need a meme generation site to go along with it...

ircbot/plugin/macros.py
dkess: Maybe we should use the same access controls here that account approval use.

chriskuehl: I'm probably a +0 for not requiring auth at this point -- the abuse potential is low and it's a pain for legitimate users (especially since Slack users currently have no way to be "trusted" by create). Maybe revisit if it becomes a problem (which I don't really think it will)?

ircbot/plugin/macros.py
```diff
@@ -0,0 +1,68 @@
...
+        if result is not None:
+            msg.respond(
+                'macro `{}` already exists as {}'.format(
+                    result['slug'],
+                    result['link'],
+                )
+            )
+        else:
+            c.execute(
+                'INSERT INTO macros (slug, link) VALUES (%s, %s)',
```
redplanks: Is this escaped?

kpengboy: Yes, this is sanitized.

ircbot/plugin/macros.py
abizer: none of the other things are "valuable contributions" are currently authenticated AFAICR, plus this is backed up and mostly logged so I'm not too concerned about malicious activity, but it shouldn't be difficult to add if it becomes a problem

abizer:
it probably wouldn't be difficult to refactor these fun commands to share a common backing table, but since this is pretty small, I figured it wasn't really a big deal. might be a cool starter project for new staff though to consolidate the plugins into a single backing table with id, time_created, trigger, response, plugin_type, is_deleted as the columns. 

there's nothing in the code really limiting it to images, it's just a straight adaptation from the equivalent feature at FB, I figured it was fine for it to be relatively specific.

PR #55, opened 2018-09-10
Add Discourse shortcut plugin

dkess:
This works exactly like the RT shortcut plugin:

```
14:13 <@dkessler> d#129
14:13 <create-dkessler> dkessler: d#129: "Kubernetes" | Projects, started by dkessler | https://ocf.io/d/129
```

### DIFF
--- a/ircbot/plugin/discourse.py
+++ b/ircbot/plugin/discourse.py
@@ -1,0 +1,21 @@
+"""Print Discourse topic information."""
+import re
+
+from ocflib.infra.discourse import DiscourseTopic
+
+
+REGEX = re.compile(r'(?:d#|ocf.io/d/)([0-9]+)')
+
+
+def register(bot):
+    bot.listen(REGEX.pattern, show_topic)
+
+
+def show_topic(bot, msg):
+    """Show Discourse topic details."""
+    for topic in REGEX.findall(msg.text):
+        try:
+            d = DiscourseTopic.from_number(bot.discourse_apikey, int(topic))
+            msg.respond(str(d))
+        except AssertionError:
+            pass
--- a/ircbot/plugin/discourse.py
+++ b/ircbot/plugin/discourse.py
@@ -1,6 +1,7 @@
 """Print Discourse topic information."""
 import re
 
+from ocflib.infra.discourse import DiscourseError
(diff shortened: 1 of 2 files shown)


ircbot/plugin/discourse.py
```diff
@@ -0,0 +1,21 @@
...
+    bot.listen(REGEX.pattern, show_topic)
+
+
+def show_topic(bot, msg):
+    """Show Discourse topic details."""
+    for topic in REGEX.findall(msg.text):
+        try:
+            d = DiscourseTopic.from_number(bot.discourse_apikey, int(topic))
+            msg.respond(str(d))
+        except AssertionError:
```
keur: What is this for?

kpengboy: `except AssertionError` sounds fundamentally wrong to me. If your program is correct, it should never throw an `AssertionError` in the first place. I think rather than perpetuating this practice, we ought to fix the code in ocflib with this broken behavior.

ircbot/plugin/discourse.py
```diff
@@ -0,0 +1,21 @@
+"""Print Discourse topic information."""
+import re
+
+from ocflib.infra.discourse import DiscourseTopic
+
+
+REGEX = re.compile(r'(?:d#|ocf.io/d/)([0-9]+)')
```
kpengboy: I would opine to anchor this regex and the corresponding one for RT to word boundaries.

ircbot/plugin/discourse.py
dkess: I agree with you, but what do you recommend I do here instead? I'll change it in ocflib.

ethanhs: An `AssertionError` should *always* cause the program to stop execution (after any needed resource cleanup). If this is being raised in ocflib, either your usage is incorrect, or the code in ocflib is not correct :)

ethanhs: In a case like this I would create my own subclass of ValueError and raise that. This is so you only catch your own exception, which avoids potentially swallowing a ValueError raised somewhere else (e.g. in a library you call), which would make debugging a lot harder.

PR #57, opened 2018-09-12
add help page listing available macros

abizer:
looks like this:

macros

### DIFF
--- a/ircbot/plugin/help.py
+++ b/ircbot/plugin/help.py
@@ -9,11 +9,17 @@ import jinja2
 def register(bot):
     threading.Thread(target=help_server, args=(bot,), daemon=True).start()
     bot.listen(r'^help$', help, require_mention=True)
+    bot.listen(r'^macros$', help_macro, require_mention=True)
 
 
 def help(bot, msg):
     """Provide a link to this help page."""
     msg.respond('https://ircbot.ocf.berkeley.edu/')
+
+
+def help_macro(bot, msg):
+    """Provide a link to the list of macros."""
+    msg.respond('https://ircbot.ocf.berkeley.edu/macros')
 
 
 def build_request_handler(bot):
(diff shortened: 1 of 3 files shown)


dkess:
Are we enforcing that these are image links? If not, we need a way to handle non-image macros.

abizer:
there's no good way to enforce that they are image links, I don't think. Whitelisting file extensions is the easiest to implement but pretty fragile (e.g. #m nattofriends -> https://yelp.fuqu.jp, which just returns an image), downloading the image and trying to do some imagemagick shit to see if it is valid seems extremely bad. Do you have any suggestions? Right now, if the links are not images, it just shows the broken image link, which doesn't actually break anything.

dkess:
Put the text in the `alt` attribute. If the image fails to load, that will show up in place of the image. See the source of https://www.ocf.berkeley.edu/~dkessler/image_error

abizer:
I already am - see macros.html#L40

ircbot/plugin/help.py
```diff
@@ -24,24 +30,32 @@ def build_request_handler(bot):
...
-                rendered = jinja_env.get_template('plugin/templates/help.html').render(
-                    plugins=sorted(plugins.items(), key=lambda p: p[0].__name__),
-                ).encode('utf8')
-
-                self.send_response(200, 'Okay')
-                self.send_header('Content-Length', len(rendered))
-                self.send_header('Content-Type', 'text/html; charset=utf-8')
+                self.render_response(
+                    'plugin/templates/help.html',
+                    plugins=sorted(plugins.items(), key=lambda p: p[0].__name__)
```
jvperrin: nit: add a trailing comma

ircbot/plugin/help.py
```diff
@@ -24,24 +30,32 @@ def build_request_handler(bot):
...
+                )
+            elif self.path == '/macros':
+                self.render_response(
+                    'plugin/templates/macros.html',
+                    macros=bot.plugins['macros'].list(bot)
```
jvperrin: nit: add a trailing comma here too

ircbot/plugin/macros.py
```diff
@@ -71,3 +71,12 @@ def delete(bot, msg):
             (slug,)
         )
         msg.respond('macro `{}` has been deleted.'.format(slug))
+
+
+def list(bot):
+    """list all macros for help page"""
```
chriskuehl: this should be a real sentence (capital first letter, end with a period) per https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

ircbot/plugin/macros.py
```diff
@@ -71,3 +71,12 @@ def delete(bot, msg):
...
+
+    with db.cursor(password=bot.mysql_password) as c:
+        c.execute('SELECT slug, link FROM macros ORDER BY slug DESC')
```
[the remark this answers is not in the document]
abizer: yeah good point I forgot to take this out from testing

ircbot/plugin/macros.py
```diff
@@ -71,3 +71,12 @@ def delete(bot, msg):
...
+        for entry in c.fetchall():
+            yield entry['slug'], entry['link']
```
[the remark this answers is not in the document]
abizer: interesting, thanks for noting this.

PR #62, opened 2018-10-03
add a plugin to show the contents of a linked tweet

### LINKED ISSUE #60: expand tweets inline
we currently have create expand rt, discourse, stackoverflow, etc. inline on IRC. considering how many tweets get linked on IRC, we may want to consider expanding them inline as well.

cg505:
Requires adding a new section to the config like so:

```
[twitter]
apikey=<"API key" from developer.twitter.com>
apisecret=<"API secret key" from developer.twitter.com>
```

I used a personal key to test but we should apply for an API key as an organization.

Closes #60.

### DIFF
--- a/ircbot/plugin/twitter.py
+++ b/ircbot/plugin/twitter.py
@@ -1,8 +1,9 @@
-"""Post the contents of linked tweets."""
-import requests
+"""Post the contents of tweets."""
 from functools import lru_cache
 
-API = 'https://api.twitter.com'
+import requests
+
+TWITTER_API = 'https://api.twitter.com'
 
 
 def register(bot):
@@ -12,7 +13,7 @@ def register(bot):
 @lru_cache(maxsize=1)
 def _get_token(apikeys):
     resp = requests.post(
-        '{}/oauth2/token'.format(API),
+        '{}/oauth2/token'.format(TWITTER_API),
         data={'grant_type': 'client_credentials'},
         auth=apikeys,
     )
@@ -27,7 +28,7 @@ def _get_tweet(apikeys, status_id, retry=True):
     bearer_token = _get_token(apikeys)
 
     resp = requests.get('{}/1.1/statuses/show.json?id={}&tweet_mode=extended'.format(
-        API,
+        TWITTER_API,
         status_id,
     ), headers={
         'Authorization': 'Bearer {}'.format(bearer_token),
@@ -62,7 +63,7 @@ def _format_tweet(tweet):
             _format_media(media, url),
         )
     contents = contents.replace('\n', ' ')
-    return '@{handle} ({realname}): {contents}'.format(
+    return '@\x02{handle}\x02 ({realname}): {contents}'.format(
         handle=tweet['user']['screen_name'],
         realname=tweet['user']['name'],
         contents=contents,
@@ -70,8 +71,7 @@ def _format_tweet(tweet):
 
 
 def show_tweet(bot, msg):
-    """Show tweet content."""
+    """Show the user and content of a linked tweet."""
     tweet = _get_tweet(bot.twitter_apikeys, msg.match.group(1))
     if tweet:
-        print(_format_tweet(tweet))
         msg.respond(_format_tweet(tweet), ping=False)
--- a/ircbot/plugin/twitter.py
+++ b/ircbot/plugin/twitter.py
@@ -6,7 +6,7 @@ bearer_token = None
 
 
 def register(bot):
-    bot.listen(r'https?://(mobile\.)?twitter\.com/[^/]+/status/([0-9]+)', show_tweet)
+    bot.listen(r'https?://(?:mobile\.|www\.|m\.)?twitter\.com/[^/]+/status/([0-9]+)', show_tweet)
 
 
 def _refresh_token(apikeys):
@@ -19,11 +19,11 @@ def _refresh_token(apikeys):
     resp.raise_for_status()
 
     authorization = resp.json()
-    assert(authorization['token_type'] == 'bearer')
+    assert authorization['token_type'] == 'bearer'
     bearer_token = authorization['access_token']
 
 
-def _tweet(apikeys, status_id):
+def _get_tweet(apikeys, status_id):
     if bearer_token is None:
         _refresh_token(apikeys)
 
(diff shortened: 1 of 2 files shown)


ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,59 @@
...
+        auth=apikeys,
+    )
+    resp.raise_for_status()
+
+    authorization = resp.json()
+    assert(authorization['token_type'] == 'bearer')
+    bearer_token = authorization['access_token']
+
+
+def _tweet(apikeys, status_id):
```
kpengboy: I would name this `_get_tweet`, sounds like you're trying to post a tweet...

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,59 @@
...
+    global bearer_token
+    resp = requests.post(
+        '{}/oauth2/token'.format(API),
+        data={'grant_type': 'client_credentials'},
+        auth=apikeys,
+    )
+    resp.raise_for_status()
+
+    authorization = resp.json()
+    assert(authorization['token_type'] == 'bearer')
```
kpengboy: Nit: our `assert`s elsewhere do not use parentheses

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,59 @@
+"""Post the contents of linked tweets."""
+import requests
+
+API = 'https://api.twitter.com'
+bearer_token = None
+
+
+def register(bot):
+    bot.listen(r'https?://(mobile\.)?twitter\.com/[^/]+/status/([0-9]+)', show_tweet)
```
kpengboy: Might as well catch `www` too while you're at it

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,59 @@
...
+    if bearer_token is None:
+        _refresh_token(apikeys)
+
+    resp = requests.get('{}/1.1/statuses/show.json?id={}&tweet_mode=extended'.format(
+        API,
+        status_id,
+    ), headers={
+        'Authorization': 'Bearer {}'.format(bearer_token),
+    })
+    resp.raise_for_status()
```
kpengboy: What happens in the case of a nonexistent tweet?

ircbot/plugin/twitter.py
kpengboy: Also, I would suggest making the first group a non-capturing group, since we don't actually need its value.

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,59 @@
...
+
+    return resp.json()
+
+
+def _format_tweet(tweet):
+    contents = tweet['full_text']
+    media = tweet['extended_entities']['media']
+    if media:
+        contents = contents.replace(
+            media[0]['url'],
```
kpengboy: Why only the first entry of `media`?

ircbot/plugin/twitter.py
cg505: didn't know you could do that, thanks

ircbot/plugin/twitter.py
cg505: I included mobile because something is actually served there, but www.twitter.com just redirects to twitter.com. I can add it though, certainly no harm.

ircbot/plugin/twitter.py
cg505: They all have the same value.

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,59 @@
...
+
+API = 'https://api.twitter.com'
+bearer_token = None
+
+
+def register(bot):
+    bot.listen(r'https?://(mobile\.)?twitter\.com/[^/]+/status/([0-9]+)', show_tweet)
+
+
+def _refresh_token(apikeys):
```
keur: Is there any benefits to doing this in the plugin directory and using a `global` when you could just request the `bearer_token` when `create` starts? Then you can pass the `bearer_token` as a parameter to `CreateBot` instead of the api tuple.

I could understand doing it this way if you needed to repeatedly refresh the token (some APIs do put time limits on tokens, annoyingly enough). But from glancing over the twitter api it appears `bearer_token`'s don't expire, but are only invalidated by the user.

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    resp.raise_for_status()
+
+    authorization = resp.json()
+    assert(authorization['token_type'] == 'bearer')
+    bearer_token = authorization['access_token']
+
+
+def _tweet(apikeys, status_id):
+    if bearer_token is None:
```
keur: Again, not a fan of this design because nothing magically sets the `bearer_token` to `None` after the initial refresh.

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,59 @@
+"""Post the contents of linked tweets."""
+import requests
+
+API = 'https://api.twitter.com'
```
keur: Would call this `TWITTER_API`

ircbot/plugin/twitter.py
chriskuehl: If the bearer_token truly doesn't expire, can you just put _that_ in the config file instead of the api key?

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,74 @@
...
+bearer_token = None
+
+
+def register(bot):
+    bot.listen(r'https?://(?:mobile\.|www\.|m\.)?twitter\.com/[^/]+/status/([0-9]+)', show_tweet)
+
+
+def _refresh_token(apikeys):
```
chriskuehl: If you stick with this approach, maybe a nicer way to write this could be:

```python
from functools import lru_cache


@lru_cache()
def _bearer_token():
    return ...
```

`lru_cache` will handle the caching for you, rather than you having to check `if bearer_token is not None` everywhere. Plus no globals (though still hidden global state, of course).

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,74 @@
...
+    media = tweet.get('extended_entities', {}).get('media')
+    if media:
+        # url is the same for all media
+        url = media[0]['url']
+        contents = contents.replace(
+            url,
+            _format_media(media, url),
+        )
+    contents = contents.replace('\n', ' ')
+    return '@{handle} ({realname}): {contents}'.format(
```
chriskuehl: as a nice to have, some coloring can really help for tweets like this (helps separate out the various components of the message)

even just bolding the handle + name bit could help

ircbot/plugin/twitter.py
```diff
@@ -0,0 +1,74 @@
...
+        realname=tweet['user']['name'],
+        contents=contents,
+    )
+
+
+def show_tweet(bot, msg):
+    """Show tweet content."""
+    tweet = _get_tweet(bot.twitter_apikeys, msg.match.group(1))
+    if tweet:
+        print(_format_tweet(tweet))
```
chriskuehl: reminder to remove before merging

ircbot/plugin/twitter.py
keur: That would work too. But if for some reason we needed to invalidate the token someone would need to manually fetch a new `bearer_token` using the API keys as opposed to just restarting `create`. Since that is an unlikely scenario putting the `bearer_token` in the config is fine.

kpengboy: I'd be slightly in favor of the status quo since it's more resilient to any bearer token invalidation that might happen. But no strong feelings. Just don't forget to record the API key and secret in an appropriate location.

cg505: Part of the problem is that I can't find any documentation on whether the bearer token will ever expire without being manually invalidated.

kpengboy: If you're not sure, you could play it safe and do what you're doing now. Twitter documentation indicates that if you request bearer tokens multiple times, you get the same one back each time until the token is invalidated.

cg505:
I've finally addressed review comments, would be nice if everyone could take another look.

cg505:
Once approved, this is ready to merge. The Twitter API credentials have been added on supernova and monsoon.

abizer:
you added the credentials in the puppet private share, right, and not by hand?

cg505:
no lol, was sort of confused where those files came from. will fix, please check

PR #63, opened 2018-10-14
add fallback error handling for all plugins

cg505:
so create doesn't crash anytime anyone says "create status"

### DIFF
--- a/ircbot/ircbot.py
+++ b/ircbot/ircbot.py
@@ -219,7 +219,14 @@ class CreateBot(irc.bot.SingleServerIRCBot):
                         nick=user,
                         respond=respond,
                     )
-                    listener.fn(self, msg)
+                    try:
+                        listener.fn(self, msg)
+                    except Exception as ex:
+                        msg.respond('Exception in {module}/{function}: {exception}'.format(
+                            module=listener.fn.__module__,
+                            function=listener.fn.__name__,
+                            exception=ex
+                        ))
 
             # everything gets logged
             self.recent_messages[event.target].appendleft((user, raw_text))
--- a/ircbot/ircbot.py
+++ b/ircbot/ircbot.py
@@ -29,8 +29,11 @@ from ircbot.plugin import rackspace_monitoring
 IRC_HOST = 'irc'
 IRC_PORT = 6697
 
+# TODO: set this value in the Dockerfile, instead of relying on this kludge
 user = getpass.getuser()
-if user == 'nobody':
+TESTING = user != 'nobody'
+
+if not TESTING:
     IRC_NICKNAME = 'create'
     IRC_CHANNELS_OPER = frozenset(('#rebuild', '#atool'))
     IRC_CHANNELS_ANNOUNCE = frozenset(('#atool',))
@@ -232,7 +235,7 @@ class CreateBot(irc.bot.SingleServerIRCBot):
                         )
                         msg.respond(error_msg, ping=False)
                         # don't send emails when running as dev
-                        if user == 'nobody':
+                        if not TESTING:
                             send_problem_report(dedent(
                                 """
                                 {error}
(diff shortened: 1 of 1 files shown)


ircbot/ircbot.py
```diff
@@ -219,7 +219,14 @@ def respond(raw_text, ping=True):
                         nick=user,
                         respond=respond,
                     )
-                    listener.fn(self, msg)
+                    try:
+                        listener.fn(self, msg)
+                    except Exception as ex:
+                        msg.respond('Exception in {module}/{function}: {exception}'.format(
```
kpengboy: I would suggest to use `ocflib.misc.mail.send_problem_report` instead.

cg505:
Since IRC is interactive, it seems useful to have some sort of immediate response. imo we should do both `msg.respond` and `ocflib.misc.mail.send_problem_report`. Thanks for pointing this out.

kpengboy:
Sure, that sounds reasonable.

ircbot/ircbot.py
```diff
@@ -219,7 +222,41 @@ def respond(raw_text, ping=True):
...
+                        listener.fn(self, msg)
+                    except Exception as ex:
+                        error_msg = 'ircbot exception in {module}/{function}: {exception}'.format(
+                            module=listener.fn.__module__,
+                            function=listener.fn.__name__,
+                            exception=ex
+                        )
+                        msg.respond(error_msg, ping=False)
+                        # don't send emails when running as dev
+                        if user == 'nobody':
```
kpengboy: This sounds like an awful way to determine when running as dev mode. Perhaps do [what ocfweb does](https://github.com/ocf/ocfweb/blob/9a24e756928a6341260c42e10069c3581a142a55/Dockerfile#L46-L48).

ircbot/ircbot.py
```diff
@@ -219,7 +222,41 @@ def respond(raw_text, ping=True):
...
                         respond=respond,
                     )
-                    listener.fn(self, msg)
+                    try:
+                        listener.fn(self, msg)
+                    except Exception as ex:
+                        error_msg = 'ircbot exception in {module}/{function}: {exception}'.format(
+                            module=listener.fn.__module__,
+                            function=listener.fn.__name__,
+                            exception=ex
```
kpengboy: Are there any cases where an exception message might happen to contain an API key? I don't think so, right?

ircbot/ircbot.py
cg505: This is what is done at the top of the file (line 30). Ideally we should change both. Should we still use the Dockerfile technique and reference that in both places?

ircbot/ircbot.py
cg505: I think it's possible, but I'm not really sure of any generic way to catch that. It would certainly be just about impossible to write some sanitization that will also work for newly-added plugins. So even if sanitization is attempted, probably a new plugin will come along and nothing will stop it from throwing a new exception with sensitive data that's not caught by the sanitizer.

tldr; not sure how to handle this or if it's worth it

ircbot/ircbot.py
kpengboy: Hmm. In that case, it would be better to make a `TESTING` flag somewhere near the top of the file and use that (cf. https://github.com/ocf/ocfweb/blob/5d0edd74f5c33d0b4878a774ae14118f1757267d/ocfweb/settings.py#L12) instead of directly comparing the user. I think for now we can say something like `TESTING = (user == "nobody")`, and perhaps switch to the Dockerfile technique in a separate PR.

ircbot/ircbot.py
kpengboy: Probably not too much of a concern right now. If we're paranoid we could just print out the exception name I guess.

PR #65, opened 2018-10-22
Fixing my previous PR

dongkyunc:
- TIL pre-commit
- Fixed my typos
- Added useful info to the README
- I suck at computers

### DIFF
--- a/ircbot/plugin/bless.py
+++ b/ircbot/plugin/bless.py
@@ -14,7 +14,7 @@ def bless(bot, msg):
         'We are all SysAdmins on this blessed day.',
         'Thou hast made us for thyself, O Server, and our heart is restless until it finds the bug in thee.',
         'The function of prayer is not to influence the printer, but rather to change the nature of the one who prints.'
-    ))
+    )))
 
 def bless_someone(bot, msg):
     """Have create thank somebody on your behalf."""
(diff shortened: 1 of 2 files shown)


ircbot/plugin/bless.py
```diff
@@ -1,21 +1,24 @@
...
+        'The function of prayer is not to influence the printer, but rather to change the nature of the one '
+        'who prints.'
     )))
 
+
 def bless_someone(bot, msg):
     """Have create thank somebody on your behalf."""
-    msg.respond('Blessed be {}, in the name of the compiler, the linker, and the holy runtime!'.format(msg.match.group(1)), ping=False)
+    msg.respond('Blessed be {}, in the name of the compiler, the linker, and the holy runtime!'.format(
+        msg.match.group(1)), ping=False)
```
chriskuehl: i would suggest formatting more in the "OCF" style (like what [black](https://github.com/ambv/black) or [add-trailing-comma](https://github.com/asottile/add-trailing-comma) would do here), e.g.

```python
msg.respond(
    'Blessed be {}, in the name of the compiler, the linker, and the holy runtime!'.format(
        msg.match.group(1),
    ),
    ping=False,
)
```
this current style creates a lot of diff churn and (imo) is less readable

we should probably just adopt either black or add-trailing-comma so we stop having to worry about this on every PR

PR #73, opened 2018-12-25
plugin: give spongebob a better caps distribution

keur:
Exponentially increase the chance of a case switch on duplicates.

No more bad spongebobs!

[Here](https://i.fluffy.cc/2fwMWNJjbsjhRmFt4fq6qJpZCzV0PWcZ.html) are some spongebobs I ran in sequence.


keur:
@dkess algorithm inspired by [this](https://github.com/nolanlum/kochira_caa/commit/9477b06f77c36313f09dd566abc97d3424b8a203). I just took a graph that I liked. The distribution looks nice with longer messages. Might need to introduce an alpha for smaller length messages, though.

PR #74, opened 2018-12-25
plugin: automatically spongebob memes

keur:


dkess:
We should be careful with having the bot talk when it's not explicitly called on, but I'd be willing to try this out. If it gets annoying I'll revert it though.

fydai:
I would add Donald J. Trump (so his tweets gets spongebob'd (assuming create is allowed to respond to itself)).

keur:
@fydai Unfortunately, `create` can't respond to itself at the moment. Nothing would stop it from responding to the spongebob'd version, ad infinitum. I added the 'J' anyway.

fydai:
Maybe anything containing https://twitter.com/realDonaldTrump should be spongbob'd rather than regularly printed instead?

PR #75, opened 2018-12-26
Add !choice

fydai:
We need to be able to make important life decisions (marriages, which final to study for, whether or not to accept a pull request) through create. This pull request adds the necessary functionality.

### DIFF
--- a/ircbot/plugin/gambling.py
+++ b/ircbot/plugin/gambling.py
@@ -1,10 +1,12 @@
 """Outsource all your decision making to create."""
 import random
+import re
 
 
 def register(bot):
     bot.listen(r'^!flip$', flip)
     bot.listen(r'^!8ball( |$)', eightball)
+    bot.listen(r'^!choose (.+)$', choose)
     bot.listen(r'roll (\d+)d(\d+)', roll, require_mention=True)
 
 
@@ -13,6 +15,16 @@ def flip(bot, msg):
     msg.respond('my quantum randomness says: {}'.format(
         random.choice(('approve', 'reject')),
     ))
+
+
+def choose(bot, msg):
+    """Make your important life choices, separated with ';', with create.
+    (e.g. !choice study for cs61a; don't study for cs61a)"""
+    choices = re.findall(r'\b([^;]+);?', msg.match.group(1))
+    if not choices:
+        msg.respond("Usage: !choice study for cs61a; don't study for cs61a")
+    else:
+        msg.respond(random.choice(choices))
 
 
 def eightball(bot, msg):


ircbot/plugin/gambling.py
```diff
@@ -15,6 +16,16 @@ def flip(bot, msg):
     ))
 
 
+def choose(bot, msg):
+    """Make your important life choices, separated with ';', with create."""
+    choices = [choice.strip() for choice in msg.match.group(1).split(';')]
+    choices = [choice for choice in choices if choice != '']
```
dkess: I'd just combine these into one line-- do `strip()` twice if necessary.

ircbot/plugin/gambling.py
```diff
@@ -15,6 +16,16 @@ def flip(bot, msg):
...
+    if not choices:
+        msg.respond('Please provide some choices')
```
dkess: Explain the syntax here

ircbot/plugin/gambling.py
abizer:
```python
choices = list(filter(None, (choice.strip() for choice in msg.match.group(1).split(';'))))
```

ethanhs: Perhaps instead of manually parsing it, use `choices = re.findall(r'\b([^;]+);?', msg.match.group(1))`? This seems much better than a deeply nested statement. Some sample output:

```
>>> re.findall(r'\b([^;]+);?', '')
[]
>>> re.findall(r'\b([^;]+);?', 'a')
['a']
>>> re.findall(r'\b([^;]+);?', 'a;b')
['a', 'b']
>>> re.findall(r'\b([^;]+);?', 'a;b;c')
['a', 'b', 'c']
>>> re.findall(r'\b([^;]+);?', 'a;b;c;')
['a', 'b', 'c']
re.findall(r'\b([^;]+);?', 'a ;b ;c ;')
['a ', 'b ', 'c ']
>>> re.findall(r'\b([^;]+);?', ' a ; b ; c ;')
['a ', 'b ', 'c ']
```

ircbot/plugin/gambling.py
ethanhs: Oh and give an example too please. Perhaps "Please provide some choices separated by semicolons. e.g. choice1;choice2;choice3"

dkess: In case the suggestions wasn't clear-- the examples should be part of this error message, so users know what to do when they see this message.

patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/ircbot/plugin/gambling.py b/ircbot/plugin/gambling.py
```
ethanhs: I don't think this is supposed to be included?

PR #76, opened 2018-12-26
Add webhook support and prometheus

dkess:
Curious on people's thoughts on this. Instead of creating a new service to process IRC webhooks, we can use create's HTTP server. This could also be used to capture GitHub webhooks.


dkess:
Hmm, Flask sounds like be a good idea. Might close this at some point and re-open with a different design approach.

fydai:
We should be able to do this now, with the new Flask setup.

PR #78, opened 2018-12-29
plugin: add !widetext

keur:


ircbot/plugin/wide.py
```diff
@@ -0,0 +1,69 @@
...
+    """乇乂丅尺卂 丅卄工匚匚"""
+    text = get_text(bot, msg)
+    if text != '':
+        msg.respond(text.translate(thicc_map), ping=False)
+
+
+def widetextify(text, width=1):
+    translated_words = [''.join(
+        [c.translate(widetext_map) + space * (width - 1) for c in word],
+    )
```
jvperrin: This could be much simpler if you just include the space in the widetext_map in the first place:

```python3
widetext_map = {i: i + 0xFEE0 for i in range(0x21, 0x7F)}
widetext_map[ord(' ')] = 0x3000  # the space character has unique mapping to become a unicode ideographic space
space = chr(0x3000)
```

then this function becomes much smaller and you don't have to split by words at all:

```python3
def widetextify(text, width=1):
    return ''.join([char.translate(widetext_map) + space * (width - 1) for char in text])
```

ircbot/plugin/wide.py
```diff
@@ -0,0 +1,69 @@
...
+        for word in text.split()]
+    seperator = space * width
+    return seperator.join(translated_words)
+
+
+def get_text(bot, msg=None):
+    text = msg.match.group(1)
```
jvperrin: This function could probably just be this if you'd like, although the part about needing a default two-value tuple is annoying and I wish [PEP 505](https://www.python.org/dev/peps/pep-0505/) were here already:

```python3
def get_text(bot, msg=None):
    return msg.match.group(1) or bot.recent_messages[msg.channel].get(0, ('', ''))[1]
```

ircbot/plugin/wide.py
```diff
@@ -0,0 +1,69 @@
...
+    bot.listen(r'^!evenwidertext(?: (.*))?', evenwidertext)
+    bot.listen(r'^!superwidetext(?: (.*))?', superwidetext)
+
+    bot.listen(r'^!extrathicc(?: (.*))?', extrathicc)
+
+
+def widetext(bot, msg):
+    """ｗｅｌｃｏｍｅ　ｔｏ　ｔｈｅ　ｏｃｆ"""
+    text = get_text(bot, msg)
+    if text != '':
```
jvperrin: This check is duplicated a bunch, I think there should be a nicer way to avoid it. Could `widetextify` just include the `msg.respond` call itself (and the fetching of the text) so that the check can just be implemented once there?

Something like this:

```python3
def widetextify(width=1):
    text = get_text(bot, msg)
    if text:
        response = ''.join([char.translate(widetext_map) + space * (width - 1) for char in text])
        msg.respond(response, ping=False)
```

Then you could even make some of these into partials:

```python3
widetext = widetextify
evenwidertext = functools.partial(widetextify, width=2)
superwidetext = functools.partial(widetextify, width=3)
```

keur: ooo so ~fancy~ i luv it

ircbot/plugin/wide.py
keur: This won't quite work, because [when you specify a channel you get a deque](https://github.com/ocf/ircbot/blob/404c21c34bb42da34bab55548671576a98857e59/ircbot/ircbot.py#L106-L108), which doesn't have a `get()` method. I still was able to clean it up in a similar manner though, since empty deques evaluate will evaluate to false in a conditional.

ircbot/plugin/wide.py
keur: This won't quite work, because [when you specify a channel you get a deque](https://github.com/ocf/ircbot/blob/404c21c34bb42da34bab55548671576a98857e59/ircbot/ircbot.py#L106-L108), which doesn't have a `get()` method. I still was able to clean it up in a similar manner though, since empty deques evaluate will evaluate to false in a conditional.

abizer:
I would have named the triggers !w, !w2 and !w3, much more user friendly than !widetext, !evenwidertext etc

PR #80, opened 2018-12-29
plugin: add functools.partial support to help

keur:


abizer:
honestly I would simplify the widetext plugin and remove the 2 using partials entirely rather than doing this but it's fine I guess

jvperrin:
So only having the single spaced option essentially?

abizer:
yeah. the naming on those triggers is pretty awkward/long to remember and use as well

jvperrin:
That's fair, I do agree it'll be hard to remember them.

I do think functools.partial support is a different issue though that would still be useful to support.

keur:
@abizer if you want to change the triggers just push it straight to master. Fine by me.

keur:
I say simplify the triggers, leave the partials, and add this?

abizer:
yeah go for it

PR #83, opened 2018-12-29
Replace tt tag with class="command"

dkess:
The `<tt>` tag is considered obsolete (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt). It can be replaced with CSS, so this commit will fix it.

### DIFF
--- a/ircbot/plugin/templates/help.html
+++ b/ircbot/plugin/templates/help.html
@@ -3,7 +3,7 @@
     <head>
         <title>ocf ircbot help</title>
         <style>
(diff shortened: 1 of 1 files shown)


chriskuehl:
It might be better to use `<code>` or something instead of relying solely on a CSS class. Just using CSS provides no semantic meaning, which is non-ideal for anything parsing the page non-visually (bots, screen readers, etc.). The MDN page you linked also suggests this.

dkess:
Done in 99e7f5b850a133e2f495de3ab3ad5b42378c9038

PR #84, opened 2018-12-29
help: make ircbot development more supernova friendly.

keur:
someone was running jupyter on 8888 on supernova, making the help page error out. let's just fix this for everyone.

### DIFF
--- a/ircbot/plugin/help.py
+++ b/ircbot/plugin/help.py
@@ -2,6 +2,7 @@
 import collections
 import functools
 import http.server
+import os
 import threading
 
 import jinja2
@@ -44,9 +45,11 @@ def build_request_handler(bot):
                 plugins = collections.defaultdict(set)
                 for listener in bot.listeners:
                     if isinstance(listener.fn, functools.partial):
-                        plugins[bot.plugins[listener.fn.func.__module__]].add(listener)
+                        module = listener.fn.func.__module__
                     else:
-                        plugins[bot.plugins[listener.fn.__module__]].add(listener)
+                        module = listener.fn.__module__
+                    plugins[bot.plugins[module]].add(listener)
+
                 self.render_response(
                     'plugin/templates/help.html',
                     plugins=sorted(plugins.items(), key=lambda p: p[0].__name__),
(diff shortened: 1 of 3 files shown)


ircbot/plugin/help.py
```diff
@@ -65,5 +68,6 @@ def do_GET(self):
 
 
 def help_server(bot):
-    server = http.server.HTTPServer(('0.0.0.0', 8888), build_request_handler(bot))
+    port = os.getenv('RANDOM_PORT') or 8888
```
dkess: Can this be a command line argument or config file option instead?

Also you should do `os.getenv('RANDOM_PORT', 8888)` anyways

keur: It could be, but I don't really see why that's better. Then we move the port generating logic inside of IRCBot.

keur: Then you also add print statements to the IRCBot about which port you are listening on for the dev site... I dunno. I don't like it.

dkess: ok I guess the environment variable is fine but I'd name it something else in the code, like `HTTP_PORT`

keur: kk... sec

ircbot/plugin/help.py
```diff
@@ -44,9 +45,11 @@ def do_GET(self):
                 plugins = collections.defaultdict(set)
                 for listener in bot.listeners:
                     if isinstance(listener.fn, functools.partial):
-                        plugins[bot.plugins[listener.fn.func.__module__]].add(listener)
+                        module = listener.fn.func.__module__
                     else:
-                        plugins[bot.plugins[listener.fn.__module__]].add(listener)
+                        module = listener.fn.__module__
```
jvperrin: Actually, after looking into listeners more, you can just make this [`listener.plugin_name`](https://github.com/ocf/ircbot/blob/2a1e41eac4dfcca73005f286e2c2748d1eaf9c50/ircbot/ircbot.py#L65-L67) which is clearer, maybe something similar should be added for partials?

Not sure how much sense that would make, but maybe `listener.partial_plugin_name` works?

jvperrin: Actually, never mind, I think the listener itself should have this logic (checking if it is a partial or not) in the `plugin_name` property, that would make a lot more sense than having a completely separate method here.

PR #87, opened 2018-12-31
plugin: add turing

keur:
creates a markov chain model from the quotes and inspire databases.
create uses this model to return... interesting mixtures of the two.

Here are some of my [favorites](https://i.fluffy.cc/S90VX4B4PftGS4BlpTwcWSj9zdV2fcxc.html). You can play with it in #test. `#create-keur: turing` will trigger it.

### DIFF
--- a/ircbot/plugin/turing.py
+++ b/ircbot/plugin/turing.py
@@ -1,5 +1,4 @@
 """Is create turing complete?"""
-import itertools
 import re
 
 import markovify
@@ -22,7 +21,7 @@ def markov(bot, msg):
     """Return the best quote ever"""
     if final_model:
         msg.respond(
-            final_model.make_sentence(tries=100),
+            final_model.make_sentence(tries=300),
             ping=False,
         )
 
@@ -43,8 +42,7 @@ def build_models(bot, msg=None):
         rants = c.fetchall()
 
     # Normalize the quote data... Get rid of IRC junk
-    normalized_quotes_2d = [normalize_quote(d['quote']) for d in quotes]
-    flat_quotes = list(itertools.chain(*normalized_quotes_2d))
+    clean_quotes = [normalize_quote(d['quote']) for d in quotes]
 
     # Normalize the inspire data... Just lightly prune authors
     clean_inspirations = [normalize_inspiration(d['text']) for d in inspirations]
@@ -56,24 +54,19 @@ def build_models(bot, msg=None):
     # More heavily weight our quotes and rants
     global final_model
     rants_model = markovify.NewlineText('\n'.join(clean_rants))
-    quotes_model = markovify.NewlineText('\n'.join(flat_quotes))
+    quotes_model = markovify.NewlineText('\n'.join(clean_quotes))
     inspire_model = markovify.NewlineText('\n'.join(clean_inspirations))
-    final_model = markovify.combine([quotes_model, rants_model, inspire_model], [2, 1.5, 1])
+    final_model = markovify.combine([quotes_model, rants_model, inspire_model], [2, 2, 0.5])
 
 
 def normalize_quote(q):
-    # Remove "keur:" from start of text
-    cleaned = re.sub(r'^\s*{}:\s*'.format(IRC_NICK_RE), '', q)
-    # Remove "<@keur>" and "<keur>" and "< keur>"
-    cleaned = re.sub(r'<\s*@?{}\s*>:?\s*'.format(IRC_NICK_RE), '', cleaned)
-    # Remove "-keur" from the end of text
-    cleaned = re.sub(r'\s*\-\s*{}'.format(IRC_NICK_RE), '', cleaned)
     # Remove timestamps
-    cleaned = re.sub(r'\[?\d{2}:\d{2}(:?:\d{2})?\]?', '', cleaned)
+    cleaned = re.sub(r'\[?\d{2}:\d{2}(:?:\d{2})?\]?', '', q)
     # Remove "\\" newline separators
-    cleaned = re.sub(r'\\\s', '', cleaned)
-    sentences = re.split(r'\.|\?|!', cleaned)
-    return [c.strip() for c in sentences if c]
+    cleaned = re.sub(r'\\\s*', '', cleaned)
+    # Trim punctuation from end of quotes
+    cleaned = re.sub(r'(\.|\?|!)$', '', cleaned)
+    return cleaned
 
 
 def normalize_inspiration(q):
--- a/ircbot/plugin/turing.py
+++ b/ircbot/plugin/turing.py
@@ -16,7 +16,7 @@ def register(bot):
 
 
 def markov(bot, msg):
-    """Return the best quote ever"""
+    """Return the best quote ever."""
     if final_model:
         # This tries to generate a sentence that doesn't "overlap", or
         # share too much similarity with seeded text.
@@ -27,9 +27,8 @@ def markov(bot, msg):
             # This is also much simpler than using crazy IRC nick regex.
             # Put it in the middle of the word since nicks are quoted
             # using "<@keur>" syntax.
-            mid = len(sentence) // 2
             msg.respond(
-                ' '.join(w[:mid] + '\u2060' + w[mid:] for w in sentence.split()),
+                ' '.join(w[:len(w) // 2] + '\u2060' + w[len(w) // 2:] for w in sentence.split()),
                 ping=False,
             )
         else:
--- a/ircbot/plugin/turing.py
+++ b/ircbot/plugin/turing.py
@@ -13,22 +13,21 @@ final_model = None
 
 def register(bot):
     bot.listen(r'turing', markov, flags=re.IGNORECASE, require_mention=True)
-    bot.listen(r'^!genmodels', initialize_markov_dataset)
+    bot.listen(r'^!genmodels', build_models)
 
-    initialize_markov_dataset(bot)
+    build_models(bot)
 
 
 def markov(bot, msg):
     """Return the best quote ever"""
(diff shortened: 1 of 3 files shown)


keur:
Was going to add a pre-generated nlp model, but to rebuild the model requires pulling in a bunch of libs, so not gonna do that. The generated quotes were only marginally better than the non-nlp quotes anyway.

dkess:
Can you insert the zero-width character inside people's nicks so this doesn't highlight people all the time?

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
+"""Is create turing complete?"""
+import re
+
+import markovify
+
+from ircbot import db
+
+final_model = None
```
abizer: caps

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def register(bot):
+    bot.listen(r'turing', markov, flags=re.IGNORECASE, require_mention=True)
+    bot.listen(r'^!genmodels', build_models)
```
abizer: I'd name this something more indicative of the plugin it belongs to, maybe like "create: regenerate markov" or something like that.

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+    """Return the best quote ever"""
+    if final_model:
+        output = final_model.make_sentence(tries=200)
+        if output:
+            # Put a zero width space in every word to prevent pings
+            # This is also much simpler than using crazy IRC nick regex.
+            # Put it in the middle of the word since nicks are quoted
+            # using "<@keur>" syntax.
+            msg.respond(
+                ' '.join([w[:len(w) // 2] + '\u2060' + w[len(w) // 2:] for w in output.split()]),
```
abizer:
```suggestion
                mid = len(w) // 2
                ' '.join(w[:mid] + '\u2060' + w[mid:] for w in output.split()),
```

you can replace the list comprehension with a generator

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+        else:
+            # This has never happened, but just in case...
+            msg.respond(
+                'Could not generate sentence. Please try again or run !genmodels',
+                ping=False,
+            )
+
+
+def build_models(bot, msg=None):
+    """Rebuild the markov models"""
```
abizer: add a period to the end of the docstring sentence

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+    with db.cursor(password=bot.mysql_password) as c:
+        # Fetch quote data
+        c.execute('SELECT quote from quotes WHERE is_deleted = 0')
```
abizer: s/from/FROM/

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+        quotes = c.fetchall()
+
+        # Fetch inspire data
+        c.execute('SELECT text from inspire')
```
abizer: ibid

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+        inspirations = c.fetchall()
+
+        # Fetch iconic FOSS rants
+        c.execute('SELECT text from rants')
```
abizer: ibid, I'd also name rename this table to something like `markov_seed` or something

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # Normalize the inspire data... Just lightly prune authors
+    clean_inspirations = [normalize_inspiration(d['text']) for d in inspirations]
+
+    # Normalize the rant data... just remove ending punctuation
+    clean_rants = [normalize_rant(d['text']) for d in rants]
+
+    # Create the three models, and combine them.
+    # More heavily weight our quotes and rants
+    global final_model
```
abizer: I'd make this function side-effect free and add a wrapper function that declares `final_model` global and assigns the output of this function to it that can also be used in genmodels, e.g.

```python
def build_model(creds, weights):
    ...

def generate_model(bot, msg=None):
    global FINAL_MODEL
    FINAL_MODEL = build_model(bot.mysql_password, (2, 2, 0.5))
```

ircbot/plugin/turing.py
jvperrin: This isn't a constant though, so I'm not sure it should be capitalized. [PEP 8](https://www.python.org/dev/peps/pep-0008/#global-variable-names) for this suggests that global variables should just be named the same as regular functions.

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+
+import markovify
+
+from ircbot import db
+
+final_model = None
+
+
+def register(bot):
+    bot.listen(r'turing', markov, flags=re.IGNORECASE, require_mention=True)
```
jvperrin: Anchor this regex so that it doesn't just match `turing` in some other context? It does require a mention at least, but I think something like `create: this is a turing test` would trigger this for instance which seems a little strange.

requirements.txt
```diff
@@ -25,6 +25,7 @@ Jinja2==2.10
 jwcrypto==0.5.0
 kombu==4.2.1
 ldap3==2.5.1
+markovify==0.7.1
```
jvperrin: Add an unpinned version of this to `requirements-minimal.txt` too. Otherwise, if you don't, the next time `make update-requirements` is run, this will get dropped from this file.

ircbot/plugin/turing.py
jvperrin: Or maybe `create: turing regenerate`/`create: turing regen` or something?

Does this ever have to actually be done to need a command?

ircbot/plugin/turing.py
jvperrin: Also, instead of this being a global, can it just be constructed in the `register` hook and then passed into the `bot.listen` statement afterwards as a partial so the `markov` function will just receive the model as an argument every time? This only really works if it doesn't need to be regenerated, but from the sounds of things, I don't feel like it should ever need to actually be regenerated.

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+    bot.listen(r'^!genmodels', build_models)
+
+    build_models(bot)
+
+
+def markov(bot, msg):
+    """Return the best quote ever"""
+    if final_model:
+        output = final_model.make_sentence(tries=200)
```
jvperrin: Can this be specified as a constant at the top of this plugin instead? I think a link to the [markovify README](https://github.com/jsvine/markovify#basic-usage) or just explaining that it tries to generate text that doesn't overlap with the existing text would also be useful to explain what this is doing.

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+            # using "<@keur>" syntax.
+            msg.respond(
+                ' '.join([w[:len(w) // 2] + '\u2060' + w[len(w) // 2:] for w in output.split()]),
+                ping=False,
+            )
+        else:
+            # This has never happened, but just in case...
+            msg.respond(
+                'Could not generate sentence. Please try again or run !genmodels',
+                ping=False,
```
jvperrin: I think a `ping` here might actually be good, because it's an error case and that'll highlight it errored to whoever requested the quote.

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,95 @@
...
+            # Put it in the middle of the word since nicks are quoted
+            # using "<@keur>" syntax.
+            msg.respond(
+                ' '.join([w[:len(w) // 2] + '\u2060' + w[len(w) // 2:] for w in output.split()]),
+                ping=False,
+            )
+        else:
+            # This has never happened, but just in case...
+            msg.respond(
+                'Could not generate sentence. Please try again or run !genmodels',
```
jvperrin: I'd remove the suggestion here to regenerate the models. If it ever does get to this stage, it just means that it couldn't generate something that didn't overlap after 200 tries, so it gave up, which seems fine, but it seems strange to me that the models themselves would be at fault in that case. Has that ever been the case when testing?

You could also just run `output = final_model.make_sentence(test_output=False)` to definitely get some kind of output here as a last resort?

ircbot/plugin/turing.py
jvperrin: Meh, I kinda like this name (since it explains the contents to some extent), although `markov_rants` would be clearer as to what model it belongs to.

ircbot/plugin/turing.py
keur: If someone adds new quotes and wants to regenerate the model without restarting create, then yes. To be fair, the model won't change very much unless they add 20+ quotes

keur:
@abizer @jvperrin made suggested changes. I did opt to keep the global because if someone dumps a bunch of quotes into the seed databases, most would want to see that output reflected without waiting for ircbot to restart. I liked abizer's idea of a side-effect free function, which addresses most of the issues with using the global imo.

ircbot/plugin/turing.py
```diff
@@ -0,0 +1,104 @@
...
+
+def register(bot):
+    bot.listen(r'^turing$', markov, flags=re.IGNORECASE, require_mention=True)
+    bot.listen(r'^turing regen(?:erate)?$', generate_model, flags=re.IGNORECASE, require_mention=True)
+
+    generate_model(bot)
+
+
+def markov(bot, msg):
+    """Return the best quote ever"""
```
jvperrin: nit: Same here with the period at the end of the docstring

PR #88, opened 2019-01-01
Make it easier to test ircbot and add more documentation

fydai:
Now, running `make dev` will use the default configuration in `ocf-ircbot.conf.example` and join `#test` rather than `#${username}`. This should make it easier to run ircbot and make it easier for others too give feedback.


Makefile
```diff
@@ -19,7 +19,7 @@ venv: vendor/venv-update requirements.txt requirements-dev.txt
 dev: export HTTP_PORT ?= $(RANDOM_PORT)
 dev: venv
 	@echo "\e[1m\e[93mRunning help on http://$(shell hostname -f ):$(RANDOM_PORT)/\e[0m"
-	venv/bin/python -m ircbot.ircbot
+	venv/bin/python -m ircbot.ircbot -c ocf-ircbot.conf.example
```
abizer: ircbot development is supposed to occur on supernova which has the ircbot config located in the correct location, this change is likely to create more work for anyone working on it as they'll now need to manually specify the prod config.

ircbot/ircbot.py
```diff
@@ -34,7 +34,7 @@
     IRC_CHANNELS_JOIN_MYSQL = True
 else:
     IRC_NICKNAME = 'create-{}'.format(user)
-    IRC_CHANNELS_OPER = IRC_CHANNELS_ANNOUNCE = frozenset(('#' + user,))
+    IRC_CHANNELS_OPER = IRC_CHANNELS_ANNOUNCE = frozenset(('#test',))
```
abizer: this should be a runtime configurable option via environment variable. when debugging things I wouldn't want the bot to join #test until I have feedback to request. the join/quit spam and general bugginess would be annoying to everyone else in the channel. this should be something like

```python
frozenset(('#test' if os.getenv('IRCBOT_TEST') else '#' + user,))
```
so to make it public you can `$ IRCBOT_TEST=yes make dev` or something like that

fydai:
I feel like that it should be easier for somebody who's not in ocfroot to run some sort of test configuration, in order to encourage more easy contributions to the OCF. Maybe there should be a few more Makefile flags, for empty configuration versus production configuration and what channels to join.

fydai:
I

PR #90, opened 2019-01-14
Add thread lock to IRC connection

dkess:
The `irc` library uses Python sockets, which aren't thread-safe. This PR adds a lock and acquires the lock for any operation that sends to the IRC server.

### DIFF
--- a/ircbot/ircbot.py
+++ b/ircbot/ircbot.py
@@ -128,6 +128,8 @@ class CreateBot(irc.bot.SingleServerIRCBot):
         self.plugins = {}
         self.extra_channels = set()  # plugins can add stuff here
 
+        self.sendmsg_lock = threading.Lock()
+
         # Register plugins before joining the server.
         self.register_plugins()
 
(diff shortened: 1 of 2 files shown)


ircbot/ircbot.py
```diff
@@ -249,28 +251,33 @@ def on_invite(self, connection, event):
...
-        else:
-            self.connection.privmsg(channel, message)
+        with self.sendmsg_lock:
+            # Find the length of the full message
+            msg_len = len('PRIVMSG {} :{}\r\n'.format(
+                channel,
+                message,
+            ).encode('utf-8'))
+
+            # The message must be split up if over the length limit
```
keur: we only really need to acquire the lock here. you aren't using the connection when getting the message length

ircbot/ircbot.py
```diff
@@ -249,28 +251,33 @@ def on_invite(self, connection, event):
...
+            if msg_len > MAX_CLIENT_MSG:
```
keur: can't we just simplify this logic and remove the `if statement`? with python slices , if we are less than `MAX_CLIENT_MSG` we will just have an array of size 1.

Then we can simplify the logic further and only acquire the lock before the loop

ircbot/ircbot.py
```diff
@@ -249,28 +251,33 @@ def on_invite(self, connection, event):
         return ircbot.plugin.channels.on_invite(self, connection, event)
 
     def bump_topic(self):
-        for channel, topic in self.topics.items():
-            def plusone(m):
-                return '{}: {}'.format(m.group(1), int(m.group(2)) + 1)
+        with self.sendmsg_lock:
+            for channel, topic in self.topics.items():
+                def plusone(m):
```
keur: nit: would just make this a lambda

PR #91, opened 2019-01-20
turing: account for -slack nicks when adding 0-length spaces

cg505:
as advertised in #test

### DIFF
--- a/ircbot/plugin/turing.py
+++ b/ircbot/plugin/turing.py
@@ -23,12 +23,17 @@ def markov(bot, msg):
         # Read more here: https://github.com/jsvine/markovify#basic-usage
         sentence = final_model.make_sentence(tries=200)
         if sentence:
-            # Put a zero width space in every word to prevent pings
+            # Put a zero width space in every word to prevent pings.
             # This is also much simpler than using crazy IRC nick regex.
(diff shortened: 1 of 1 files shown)


ircbot/plugin/turing.py
```diff
@@ -23,12 +23,17 @@ def markov(bot, msg):
         # Read more here: https://github.com/jsvine/markovify#basic-usage
         sentence = final_model.make_sentence(tries=200)
         if sentence:
-            # Put a zero width space in every word to prevent pings
+            # Put a zero width space in every word to prevent pings.
             # This is also much simpler than using crazy IRC nick regex.
             # Put it in the middle of the word since nicks are quoted
-            # using "<@keur>" syntax.
+            # using "<@keur>" syntax.  Additionally, remove any -slack at
+            # the end of a nick, to avoid inserting a space like
```
dkess: Why use the middle? I think it would be better to find where the nick actually starts, and put it after the first letter.

cg505: This is probably a better approach, but it's notably harder. It's not that easy to break the message into each discrete nick.

cg505:
I'm going to merge this. I opened an issue (#100) for handling better 0-length spaces in nicks in general.

PR #92, opened 2019-01-21
Add a quotes page

### LINKED ISSUE #147: Add a quotes page
Exported from https://ocf.io/rt/6832

jvperrin:
This is what the page looks like (well the top part anyway). I'd prefer to put a screenshot here than a live page link to my dev version when it might not be running and the link will very quickly get stale anyway:

quotes-page

[rt#6832](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=6832) is what inspired this, well that and me wanting a quotes page instead of trying to remember part of a quote and having to search for it...

I'd like to split out the web server from the help plugin and put it in its own plugin (and add hooks for other plugins to use it hopefully), but I think that can come in a future PR, it's not really related to this change at all.

Closes #147.

### DIFF
--- a/ircbot/plugin/templates/quotes.html
+++ b/ircbot/plugin/templates/quotes.html
@@ -1,0 +1,35 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <title>ocf ircbot quotes</title>
+        <style>
+          table {
+            border-collapse: collapse;
+          }
+          th, td {
+            border: 1px solid black;
+            padding: 4px;
+          }
+        </style>
+    </head>
+    <body>
+        <h1>ircbot quotes: <em>stupid stuff the OCF says</em></h1>
+        <hr />
+        <table>
+          <tbody>
+            <tr>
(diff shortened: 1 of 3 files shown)


ircbot/plugin/templates/quotes.html
```diff
@@ -0,0 +1,35 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <title>ocf ircbot quotes</title>
```
kpengboy: meta charset?

abizer:
I don't think it's a good idea for this to be open to the public, especially since many of us use our real names as our nicks. admittedly a lot of this can be found on moffle or something, but I'm still hesitant.

fydai:
We do have a proper webserver setup now, so a password wall could be implemented in PAM.

jvperrin:
I'll close this, if it does actually end up happening it'll need password auth of some kind and to be reformatted to fit with the current flask web server.

PR #93, opened 2019-01-24
Corrected message splitting

Porrster:
Using the widetext command revealed a bug in the way ircbot splits messages longer than the max message length. Previously, messages were split based on the number of characters until reaching the max length. This fails when using unicode characters that are larger than one byte. Now ircbot counts the number of bytes of each character up until the max length and splits based on that.

### DIFF
--- a/ircbot/ircbot.py
+++ b/ircbot/ircbot.py
@@ -264,8 +264,20 @@ class CreateBot(irc.bot.SingleServerIRCBot):
         # The message must be split up if over the length limit
         if msg_len > MAX_CLIENT_MSG:
             # Split up the full message into chunks to send
-            msg_range = range(0, len(message), MAX_CLIENT_MSG)
-            messages = [message[i:i + MAX_CLIENT_MSG] for i in msg_range]
+            messages, msg = [], ''
+            count = 0
+
+            for char in message:
+                char_len = len(char.encode('utf-8'))
+                if count + char_len > MAX_CLIENT_MSG:
+                    messages.append(msg)
+                    count, msg = 0, ''
+
+                msg += char
+                count += char_len
+
+            if len(msg) != 0:
+                messages.append(msg)
 
             for msg in messages:
                 self.connection.privmsg(channel, msg)


ircbot/ircbot.py
```diff
@@ -264,8 +264,20 @@ def say(self, channel, message):
         # The message must be split up if over the length limit
         if msg_len > MAX_CLIENT_MSG:
             # Split up the full message into chunks to send
-            msg_range = range(0, len(message), MAX_CLIENT_MSG)
-            messages = [message[i:i + MAX_CLIENT_MSG] for i in msg_range]
+            messages, msg = [], ''
+            count = 0
+
+            for char in message:
+                char_len = len(char.encode('utf-8'))
```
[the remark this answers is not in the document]
kpengboy: Is it possible to encode the message as utf-8, then split the encoded bytestring at character boundaries so each piece is as close to the limit as possible?

ircbot/ircbot.py
```diff
@@ -264,8 +264,20 @@ def say(self, channel, message):
         # The message must be split up if over the length limit
         if msg_len > MAX_CLIENT_MSG:
             # Split up the full message into chunks to send
-            msg_range = range(0, len(message), MAX_CLIENT_MSG)
-            messages = [message[i:i + MAX_CLIENT_MSG] for i in msg_range]
+            messages, msg = [], ''
+            count = 0
+
+            for char in message:
```
kpengboy: Is it possible to encode the message as utf-8, then split the encoded bytestring at character boundaries so each piece is as close to the limit as possible?

cg505: isn't that what this does?

dkess: This proposal would potentially yield invalid UTF-8, in case we split within a codepoint. wporr's solution avoids this.

ircbot/ircbot.py
```diff
@@ -264,8 +264,20 @@ def say(self, channel, message):
         # The message must be split up if over the length limit
         if msg_len > MAX_CLIENT_MSG:
             # Split up the full message into chunks to send
-            msg_range = range(0, len(message), MAX_CLIENT_MSG)
-            messages = [message[i:i + MAX_CLIENT_MSG] for i in msg_range]
+            messages, msg = [], ''
```
cg505: can we avoid having two different variables named `msg` and `message`? Maybe `next_split_msg` or something?

ircbot/ircbot.py
```diff
@@ -264,8 +264,20 @@ def say(self, channel, message):
...
+            count = 0
+
+            for char in message:
+                char_len = len(char.encode('utf-8'))
+                if count + char_len > MAX_CLIENT_MSG:
```
cg505: I think it would be cleaner to remove the `count` and `char_len` variables and just do `if len((msg + char).encode('utf-8')) > MAX_CLIENT_MSG:` here.

ircbot/ircbot.py
kpengboy: Sorry, I think I haven't explained myself clearly.

My proposal is, instead of iterating codepoint by codepoint in a Python loop (slow!), the following algorithm be used instead:

1. Encode message as UTF-8 bytes. Call this `msg_bytes`.
2. If `msg_bytes[MAX_CLIENT_MSG - 1]` is an ASCII character, then chop it there.
3. Otherwise, walk backwards to find the end of the last code point before `msg_bytes[MAX_CLIENT_MSG - 1]` and chop it there
4. Repeat with the remainder of the bytestring.

To do better, one could additionally add detection for combining characters as well.

cg505: Unless I am mistaken, your proposal (though different) will yield the same result. Your way is indeed faster: but that speed improvement seems irrelevant since it is surely dwarfed by the follow (blocking!) TCP socket write. I prefer the current approach because it is simpler to understand and debug.

That said, I don't feel strongly at all about anything above and I don't really care exactly which of these two perfectly fine approaches are used.

Porrster:
The function is not used anywhere else and is quite concise so I figured it would be cleaner to include inside the function that calls it.

dkess:
The function should still be in outer scope, you shouldn't put functions in inner scope unless it's actually needed for something.

dkess:
It shouldn't be in the class. It doesn't need access to `self` or `cls`. Put it in global scope (no indentation).

Issue #56, opened 2018-09-12
consolidate backing tables of certain plugins


abizer:
since many plugins are quite similar and have similar backing tables, it might make sense to consolidate them into a single table. For example, quotes and inspire, what_is and macros, etc. etc. might be a decent starter ticket project for new staff.

PR #94, opened 2019-01-28
ircbot: add shorturl plugin

abizer:
let the bikeshedding begin

this basically does s/macro/shorturl/ on macros.py to build a database-backed, written-in-python, lives-on-marathon, accessible-through-ircbot, avoids-username-collisions simple redirect system for shorturls. as far as I remember, those conditions address every criticism of the previous approaches (https://github.com/ocf/puppet/pull/253). this is designed to be used along with a formal puppet-based shorturl rule that looks something like 
```puppet
redirect_rule => '^s(/.*)?* https://ircbot.ocf.berkeley.edu/shorturls$1 [R]'
```
e.g. Apache on death returns a 302 to ircbot which returns a 302 to the real destination.

perhaps a new staffer can work on applying #56 to this along with the other plugins


dkess:
Should this be an ircbot plugin? I think it would be better to have a command line or web interface for this. Or perhaps it should be part of `etc`.

abizer:
it's a database, so we can still easily build a command line or web interface for this should someone want to. I thought about putting it in `etc` but that means writing more scripts to generate an Apache config on death and this was much simpler and easier to implement.

ircbot/plugin/shorturls.py
```diff
@@ -0,0 +1,130 @@
+"""Control ocfweb shorturls through ircbot."""
+from ircbot import db
+
+KEYWORDS = {'add', 'delete', 'rename', 'replace'}
+
+
+def register(bot):
+    # [!-~] is all printable ascii except spaces
```
kpengboy: Do we want shorturls to be able to contain `?`, `&`, or `=`?

abizer:
I'm not particularly a fan of the redirector living in ircbot either - certainly, in the long term, we are overloading ircbot with functionality that ought to live elsewhere. However, right now, here's the nice thing about this approach: transitioning to a dedicated shorturl redirect service is _trivial_. I'll write it myself when Kubernetes is ready. All we'd have to do is change the rule in puppet from redirect `/s/<whatever>` to `ircbot.ocf.berkeley.edu/shorturls/` to `shorturls.ocf.berkeley.edu/<whatever>` and register the new redirect bouncer at that address (then delete a few lines of code in this repo). 

Considering the length of time over which we've repeatedly bikeshed **the right** way to engineer this dead simple system, if we ever want to actually implement this feature, I think we need to get something up that works and meets our needs and iterate on it to improve it rather than relegating it to development hell.

We already have some services working in dev on k8s - templates, mastodon. keur even has a provisional ocfweb working on k8s. It shouldn't be long until we have prod services running on k8s, at which point I'll transition the redirector. I'm willing to live with suboptimal design for a short period of time in order to get the feature up so we can make use of it this semester rather than continuing to wait and invariably ending up with the feature once again not being implemented.

ircbot/plugin/shorturls.py
```diff
@@ -0,0 +1,130 @@
...
+    if new_slug in KEYWORDS:
+        msg.respond('`{}` is a reserved keyword.'.format(new_slug))
+        return
+
+    with db.cursor(password=bot.mysql_password) as c:
+        c.execute(
+            'UPDATE shorturls SET slug = %s WHERE slug = %s',
+            (new_slug, old_slug),
+        )
+        msg.respond('shorturl `{}` has been renamed to `{}`'.format(old_slug, new_slug))
```
cg505: what happens when the old slug does not exist?

ircbot/plugin/shorturls.py
```diff
@@ -0,0 +1,130 @@
...
+
+    slug = msg.match.group(1)
+    new_target = msg.match.group(2)
+
+    with db.cursor(password=bot.mysql_password) as c:
+        c.execute(
+            'UPDATE shorturls SET target = %s WHERE slug = %s',
+            (new_target, slug),
+        )
+        msg.respond('shorturl `{}` updated'.format(slug))
```
cg505: likewise, what happens when the slug does not exist?

ircbot/plugin/shorturls.py
abizer: nothing, the where clause doesn't match anything so it's a no-op

ircbot/plugin/shorturls.py
abizer: ibid

ircbot/plugin/help.py
```diff
@@ -10,7 +10,8 @@
 def register(bot):
     threading.Thread(target=help_server, args=(bot,), daemon=True).start()
     bot.listen(r'^help$', help, require_mention=True)
-    bot.listen(r'^macros$', help_macro, require_mention=True)
+    bot.listen(r'^macros?$', help_macro, require_mention=True)
+    bot.listen(r'^shorturls?$', help_shorturls, require_mention=True)
```
kpengboy: Do we plan to have any authentication before deleting shorturls?

kpengboy:
@abizer if it were up to me I'd write the dedicated shortlink service now and run it on Marathon if Kubernetes isn't ready yet. My two cents...

ircbot/plugin/shorturls.py
kpengboy: ...Well now this won't allow hyphens or underscores. Methinks those are still useful characters.

chriskuehl:
I gotta +1 everyone else, I think serving redirects from the ircbot is just a bit too crazy.

What if you instead configure death to proxy the shorturls to some PHP script hosted in `~staff/public_html`? It'll be a 5-line PHP script to query a row from the database and redirect to it, and you can worry about making it into a proper service with version control and all the other goodies later on.

abizer:
putting the bouncer in ocfweb seems like a good idea

ircbot/plugin/help.py
Baisang: If possible, I would try to limit deletion to ops or half-ops (do we even have half-ops?)

ircbot/plugin/shorturls.py
```diff
@@ -0,0 +1,130 @@
...
+
+    slug = msg.match.group(1)
+    target = msg.match.group(2)
+
+    if slug in KEYWORDS:
+        msg.respond('`{}` is a reserved keyword.'.format(slug))
+        return
+
+    if len(slug) > 255:
+        msg.respond('shorturl slugs must be <= 255 characters')
```
Baisang: this unfairly discriminates against all of those massive facebook URLs with a million query parameters! how else will Big ZUCC be able to do the machine learnings to see how people are clicking the links? /s

Baisang:
> Is there a record of the DB schema anywhere? Is that something we do?

I do love me some schematized data, we can at least store some random .sql file somewhere in git

ircbot/plugin/help.py
Baisang: same with replacement and similar non-adding commands

ircbot/plugin/shorturls.py
abizer: this only limits the size of the slug itself, not the target

kpengboy:
Thanks again for taking the initiative on this. I approve of the basic structure of this as it now stands; I still think there are minor points that still ought to be addressed, which I've commented on.

PR #95, opened 2019-02-02
Adds ircbot command to ping staff in lab

NotRyan:
Adds a command to ping all staff in the lab. 

Currently bound to '@labstaff', which will trigger Create to repeat the message, with your name at the front, and a list of labstaff's nicks where you wrote '@labstaff'. An example can be seen below.
screenshot_2019-02-01_17-39-43

Suggest any formatting changes here.

Note that this does not currently do a proper slack ping. Normally when a user in irc types out a 'xxx-slack' username, its converted to a @xxx on Slack, but that doesn't seem to happen when a bot types it. That's something to change with the slackbridge though, not this.

### DIFF
--- a/ircbot/plugin/lab.py
+++ b/ircbot/plugin/lab.py
@@ -6,7 +6,7 @@ from ocflib.lab.stats import users_in_lab_count
 def register(bot):
     bot.listen(r'is ([a-z]+) in the lab', in_lab, require_mention=True)
     bot.listen(r"(who is|who's) in the lab", who_is_in_lab, require_mention=True)
-
+    bot.listen(r"@labstaff", ping_lab_staff, require_mention=False)
 
 def in_lab(bot, msg):
     """Check if a staffer is in the lab."""
@@ -44,3 +44,15 @@ def who_is_in_lab(bot, msg):
         len(staff),
         staff_list,
     ))
+
+def ping_lab_staff(bot, msg):
+    """Ping everyone who is currently in the lab."""
+    staff = {session.user for session in staff_in_lab()}
+    channel_users = bot.channels[msg.channel].users()
+    staff_nicks = set() 
+
+    for staff_username in staff:
+        staff_nicks = staff_nicks | {nick for nick in channel_users if nick.startswith(staff_username)}
+    nicks_string = ', '.join(sorted(staff_nicks))
+
+    msg.respond(msg.text.replace("@labstaff", nicks_string)) 


dkess:
I am not sure this could work, for two reasons:

* It's a bit invasive to ping everyone like this. However I block mentions from `create` anyways so if everyone else is fine with it then it's ok.
* Slack/IRC names don't necessarily correspond to OCF usernames, so this will miss certain people.

abizer:
we did turn this off for `create: who is in the lab` (https://github.com/ocf/ircbot/commit/bccb5ba6bc8094b3aebdb34860b642fa0a70b79f#diff-ba569ccb3165ef4ea2953acbbbb00e72) on purpose

kpengboy:
> It's a bit invasive to ping everyone like this

Eh...IMO sometimes it's useful, if an issue does require attention from someone physically present in the lab...

abizer:
call it 'weewooweewoo' :P

NotRyan:
Pinging with `create: who is in the lab` is spam because that command is used just see who is in the lab. Pinging them would just be a side effect of that and serves no purpose and ends up being spam because its an incidental ping that would happen needlessly. This, on the other hand, is a dedicated ping button. There is no reason that anyone would need to use this other than to ping people. It would only end up as spam if someone chooses to spam it.

As for the "invasiveness", I don't see how this can be considered invasive at all. It's pinging people by a public mention of their nick in a public irc channel. Any user can achieve the same exact thing by typing out a name. If being contacted through this is a serious concern, there's always the option of blocking mentions from `create` as dkess brought up.

ircbot/plugin/lab.py
```diff
@@ -44,3 +44,15 @@ def who_is_in_lab(bot, msg):
...
+    """Ping everyone who is currently in the lab."""
+    staff = {session.user for session in staff_in_lab()}
+    channel_users = bot.channels[msg.channel].users()
+    staff_nicks = set() 
+
+    for staff_username in staff:
+        staff_nicks = staff_nicks | {nick for nick in channel_users if nick.startswith(staff_username)}
+    nicks_string = ', '.join(sorted(staff_nicks))
+
+    msg.respond(msg.text.replace("@labstaff", nicks_string))
```
keur: why not just send the `nicks_string` instead of returning the replacement?

NotRyan: My reasoning was that so people being pinged could see the actual message that the ping was for in the notification itself, so phone users could see it without having to actually open it up. However, its true that it does bloat things a bit.

PR #96, opened 2019-02-05
plugin: add scramble

yvbbrjdr:
A plugin that scrambles words in a sentence.

### DIFF
--- a/ircbot/plugin/scramble.py
+++ b/ircbot/plugin/scramble.py
@@ -1,0 +1,35 @@
+"""Sblmraec yruo setenecn"""
+import random
+import string
+
+
+def register(bot):
+    bot.listen(r'^!scramble(?: (.*))?', scramble)
+
+
+def scramble(bot, msg):
+    text = msg.match.group(1)
+    if text is None:
+        if len(bot.recent_messages[msg.channel]) == 0:
+            return
+        _, text = bot.recent_messages[msg.channel][0]
+    msg.respond(scramble_sentence(text), ping=False)
+
+
+def scramble_sentence(text):
+    ret = word = ''
+    for c in text:
+        if c.isspace() or c in string.punctuation:
+            ret += scramble_word(word) + c
+            word = ''
+        else:
+            word += c
+    return ret + scramble_word(word)
+
+
+def scramble_word(word):
+    if len(word) > 0:
+        tmp = list(word[1:])
+        random.shuffle(tmp)
+        return word[0] + ''.join(tmp)
+    return ''


yvbbrjdr:
I didn't use race condition to generate randomness because that's more complex than this one.

ircbot/plugin/scramble.py
```diff
@@ -0,0 +1,32 @@
...
+def scramble_sentence(text):
+    ret = ''
+    word = ''
+    for c in text:
+        if c.isspace() or c in string.punctuation:
+            ret += scramble_word(word) + c
+            word = ''
+        else:
+            word += c
+    return ret + scramble_word(word)
```
dkess: I would do a string split on the sentence, and then run scramble on each element of that list, then re-join the list of strings.

yvbbrjdr: The reason I did this is I want to treat punctuations the same as white spaces. But yeah, a regex may solve this problem.

yvbbrjdr: If you split and rejoin, you might lose some information, such as consecutive spaces.

yvbbrjdr: You might also scramble the punctuations into the word.

ircbot/plugin/scramble.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def scramble_word(word):
+    if len(word) > 0:
+        tmp = list(word[1:])
+        random.shuffle(tmp)
+        return word[0] + ''.join(tmp)
```
abizer: you can inline the temporary variable - ```return word[0] + ''.join(random.shuffle(list(word[1:])))```

ircbot/plugin/scramble.py
```diff
@@ -0,0 +1,36 @@
...
+    bot.listen(r'^!scramble(?: (.*))?', scramble)
+
+
+def scramble(bot, msg):
+    text = msg.match.group(1)
+    if text is None:
+        if len(bot.recent_messages[msg.channel]) == 0:
+            return
+        _, text = bot.recent_messages[msg.channel][0]
+    msg.respond(scramble_sentence(text), ping=False)
```
abizer: rather than an empty return, perhaps just invert the condition?
```python
if len(bot.recent_messages[msg.channel]) != 0:
  _, text = bot.recent_messages[msg.channel][0]
  msg.respond(scramble_sentence(text), ping=False)
```

ircbot/plugin/scramble.py
yvbbrjdr: `random.shuffle` doesn't return.

abizer: ah, you're right.

ircbot/plugin/scramble.py
keur: No, you definitely don't want to use split. If someone uses multiple spaces that won't be preserved on re-join. Also, the case for handling punctuation makes the logic more convoluted.

PR #97, opened 2019-02-06
add friendly support for our celsius readers

MCer4294967296:
The title says everything.

### DIFF
--- a/ircbot/plugin/weather.py
+++ b/ircbot/plugin/weather.py
@@ -6,13 +6,13 @@ import requests
 
 
 def register(bot):
-    bot.listen(r'^(?:weather|hot|cold)(-c)? ?(.*)$', weather, require_mention=True)
+    bot.listen(r'^(?:weather|hot|cold)\s*(-c)?\s?(.*)$', weather, require_mention=True)
 
 
 def weather(bot, msg):
     """Show weather for a location (defaults to Berkeley)."""
-    unit = 'celsius' if msg.match.group(1) else 'fahrenheit'
-    where = msg.match.group(2) or 'Berkeley, CA'
+    unit = 'c' if msg.match.group(1) else 'f'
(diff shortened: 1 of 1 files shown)


ircbot/plugin/weather.py
```diff
@@ -6,16 +6,17 @@
 
 
 def register(bot):
-    bot.listen(r'^(?:weather|hot|cold) ?(.*)$', weather, require_mention=True)
+    bot.listen(r'^(?:weather|hot|cold)(-c)? ?(.*)$', weather, require_mention=True)
```
dkess: Is this supposed to be in the style of command line flags? If so there should be a space here.

MCer4294967296: So I previously wanted to do it flag-style, but I was not sure about the spacing here (that's a question mark after the space, minor concerns). Therefore I later decided that it's acceptable to put -c right after the prompt, so no space.

dkess: You should be able to handle spacing by using the whitespace wildcard (`\s+`)

MCer4294967296: Then what do you think about the ` ?` after that? Is it also better to use `\s+`?

dkess: `r'^(?:weather|hot|cold)(\s+-c)?(\s+.*)$'` should be fine. I assume the weather API will strip trailing and leading whitespace from the location.

MCer4294967296:
Okay there actually are some more work in terms of color and the unit that appears in the message.

MCer4294967296:
I am having a hard time deciding how to describe the functionality of `-c` on the help page. Some help plz?

dkess:
The following options are what I came up with:

* Add a sentence to the help text explaining that `-c` uses celsius
* Make it a different command entirely and document it as such, in a manner similar to widetext (where each degree of widening is in its own bullet point on the help page)

PR #98, opened 2019-02-08
plugin: add pipe

yvbbrjdr:
Pipe the output of a command to the input of the next command.

Signed-off-by: Yuan Zhou <jerryzhou@ocf.berkeley.edu>

### DIFF
--- a/ircbot/plugin/pipe.py
+++ b/ircbot/plugin/pipe.py
@@ -1,0 +1,69 @@
+"""Pipe the output of a command to the input of the next command."""
+from ircbot.ircbot import MatchedMessage
+
+
+MAX_LENGTH = 5000
+
+
+def register(bot):
+    bot.listen(r'^!pipe (.*)$', pipe)
+
+
+def pipe(bot, msg):
+    """Pipe the output of a command to the input of the next command."""
+    commands = filter(None, [s.strip() for s in msg.match.group(1).split('|')])
+    stream = ''
+    for command in commands:
+        if stream:
+            command += ' ' + stream
+        try:
+            stream = run_command(command, bot, msg)
+        except Exception as ex:
+            msg.respond(str(ex))
+            return
+    length = len(stream)
+    if length > MAX_LENGTH:
+        msg.respond('Message length limit exceeded: {} > {}'.format(
+            length,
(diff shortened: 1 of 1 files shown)


ircbot/plugin/pipe.py
```diff
@@ -0,0 +1,66 @@
...
+
+def register(bot):
+    bot.listen(r'^!pipe (.*)', pipe)
+
+
+def pipe(bot, msg):
+    """Pipe the output of a command to the input of the next command."""
+    try:
+        commands = shlex.split(msg.match.group(1))
+    except Exception as ex:
```
cg505: I think we should specifically only catch `ValueError` here.

ircbot/plugin/pipe.py
```diff
@@ -0,0 +1,66 @@
...
+                listener.fn(bot, stub_msg)
+            except Exception as ex:
+                msg.respond(
+                    'Command {} not supported: {}'.format(
+                        listener.plugin_name,
+                        ex,
+                    ), ping=False,
+                )
+                return
+            break
```
cg505: We probably want to consider multiple matches. For instance, this is why "who is in the lab" doesn't work. It matches the "what_is" plugin. Maybe we should take the longest response.

ircbot/plugin/pipe.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def pipe(bot, msg):
+    """Pipe the output of a command to the input of the next command."""
+    try:
+        commands = shlex.split(msg.match.group(1))
+    except Exception as ex:
+        msg.respond(str(ex), ping=False)
+        return
+    stream = ''
```
cg505: We may want to consider the ability to start with something in the stream. Consider something like `test !pipe !spongebob !w` instead of `!pipe "!spongebob test" !w` which is kind of annoying.

yvbbrjdr: Yes. I was thinking about a more intuitive syntax, such as `!pipe !spongebob test | !w`. What do you think?

cg505: I'm down for that, although that sounds substantially harder to parse.

PR #101, opened 2019-02-12
plugin: update pipe

yvbbrjdr:
Add !repeat command

Signed-off-by: Yuan Zhou <jerryzhou@ocf.berkeley.edu>

This feature is brought up by @dongkyunc. I think it is fun so I implemented it, but it can well cause spam in the channels. So this PR is to decide whether we would like to add this.

Example:
```
!repeat 5 thank jerryzhou
thanks, thanks, thanks, thanks, thanks, jerryzhou!!!!!
```

### DIFF
--- a/ircbot/plugin/pipe.py
+++ b/ircbot/plugin/pipe.py
@@ -7,6 +7,7 @@ MAX_LENGTH = 5000
 
 def register(bot):
     bot.listen(r'^!pipe (.*)$', pipe)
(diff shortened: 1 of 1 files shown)


keur:
I'd recommend a much lower upper bound

dkess:
Can you give another example of what this would be used for? Most of our current transformations aren't really meaningful when applied multiple times

PR #108, opened 2019-03-14
create => python3.7

asottile:


dkess:
Pre commit config also needs to be changed to specify 3.7 ;)

PR #109, opened 2019-03-14
Add mypy checking to ircbot

ethanhs:
This is the minimal change set needed to have mypy start checking ircbot. Hopefully we can start annotating more functions and improve type checking coverage.


cg505:
gotta get the free advertising in 👌👌👌

PR #116, opened 2019-03-19
Mentioning "create" is now case insensitive

### LINKED ISSUE #115: Use case-insensitive matching for commnads that require mention
For example, if someone writes:

```
Create: who is in the lab
```

This won't work, since ircbot requires a case-sensitive matching. Sometimes people capitalize the name, so we should still support this case.

NotRyan:
You can now yell at CREATE. No more having to retype your message after your phone autocapitalizes the first letter.

fixes #115

### DIFF
--- a/ircbot/ircbot.py
+++ b/ircbot/ircbot.py
@@ -53,9 +53,6 @@ DSA_FREQ = 5
 # amount of text, so cut into small blocks to avoid that.
 MAX_CLIENT_MSG = 435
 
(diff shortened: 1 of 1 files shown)


dkess:
What is the purpose of the `MENTIONS_REQUIRE_CASE` constant? This isn't an option that'll get turned on and off, it should just always be enabled.

NotRyan:
Fair enough, I'll change that

PR #117, opened 2019-03-27
add reverse fun-ctionality

MCer4294967296:
Because I feel spammy so feel free to kill this pr :3

### DIFF
--- a/ircbot/plugin/reverse.py
+++ b/ircbot/plugin/reverse.py
@@ -1,0 +1,19 @@
+"""Scramble the words in your sentence in a readable way."""
+
+
+def register(bot):
+    bot.listen(r'^!reverse(?: (.*))?', reverse)
+
+
+def reverse(bot, msg):
+    """Reverse the sentence."""
+    text = msg.match.group(1)
+    if text is None:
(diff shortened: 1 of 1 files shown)


ircbot/plugin/reverse.py
```diff
@@ -0,0 +1,19 @@
+"""Scramble the words in your sentence in a readable way."""
```
yvbbrjdr: Please change this docstring.

ircbot/plugin/reverse.py
```diff
@@ -0,0 +1,19 @@
...
+def reverse(bot, msg):
+    """Reverse the sentence."""
+    text = msg.match.group(1)
+    if text is None:
+        if len(bot.recent_messages[msg.channel]) == 0:
+            return
+        _, text = bot.recent_messages[msg.channel][0]
+    newText = ''
+    length = len(text)
+    for i in range(length):
```
yvbbrjdr: use `reverse` builtin rather than for loop

MCer4294967296:
ye, that's indeed more superior.

Issue #106, opened 2019-03-10
add kanboard integration


cg505:
Like d#999 or rt#9999, k#99 should link to ocf.io/k/99 and provide a short description of the task.

dkess:
If someone wants to work on this independently of the ocflib issue, we could initially make it just convert the link to "ocf.io/k/99". Once the ocflib issue is done we can actually have it show the task title

dkess:
A more comprehensive integration could be automatically adding an "external link" to the IRC logs every time a Kanboard ticket is mentioned

AndrewKe:
Working on this

willtunnels:
@AndrewKe just fyi, a commit got merged into ocflib to add kanboard support

cg505:
#113 implemented the first iteration of this functionality, but we should still use the new ocflib functions and display more detailed support.

cg505:
closed by #106

PR #118, opened 2019-04-05
display more info about Kanboard tasks (#106)

willtunnels:
This should work. The only problem is that querying the KanboardTask is 401'ing. My suspicion is that the apikey in /etc/ocf-ircbot/ocf-ircbot.conf isn't right but I can't really check that since I'm not an admin on Kanboard.


dkess:
Ah, I see why this wasn't working. The API key is for the "gstaff" user, so you need to specify that as the username. I'll make a quick commit for that right now.

PR #126, opened 2019-04-23
github: catch and ignore NotFoundErrors

dkess:
If someone types an invalid GitHub repository name, it triggers an exception and emails rootspam. This PR fixes that behavior by silently ignoring the error.

### DIFF
--- a/ircbot/plugin/github.py
+++ b/ircbot/plugin/github.py
@@ -13,9 +13,8 @@ def project_info(bot, msg):
     user, repo_name = msg.match.groups()
 
     github = github3.GitHub()
-    repo = github.repository(user, repo_name)
-
-    if repo is not None:
+    try:
+        repo = github.repository(user, repo_name)
         msg.respond(
             '\x0303{user}/{repo}\x03 | \x0308{stars} stars\x03 | \x0314{description}\x03'.format(
                 user=user,
@@ -25,6 +24,8 @@ def project_info(bot, msg):
             ),
             ping=False,
         )
+    except github3.exceptions.NotFoundError:
+        pass
 
 
 def issue_info(bot, msg):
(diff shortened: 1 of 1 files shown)


ircbot/plugin/github.py
```diff
@@ -32,11 +35,16 @@ def issue_info(bot, msg):
...
-    if repo is not None:
-        issue = repo.issue(int(issue_num))
-        if issue is not None:
+    try:
+        repo = github.repository(user, repo_name)
+    except github3.exceptions.NotFoundError:
+        pass
+    else:
+        try:
+            issue = repo.issue(int(issue_num))
```
cg505: why don't we just have one try/except block that gets both the repo and the issue?

ircbot/plugin/github.py
```diff
@@ -53,11 +61,16 @@ def pr_info(bot, msg):
...
-    if repo is not None:
-        pr = repo.pull_request(int(pr_num))
-        if pr is not None:
+    try:
+        repo = github.repository(user, repo_name)
+    except github3.exceptions.NotFoundError:
+        pass
+    else:
+        try:
+            pr = repo.pull_request(int(pr_num))
```
cg505: same here

ircbot/plugin/github.py
```diff
@@ -13,7 +13,10 @@ def project_info(bot, msg):
     user, repo_name = msg.match.groups()
 
     github = github3.GitHub()
-    repo = github.repository(user, repo_name)
+    try:
+        repo = github.repository(user, repo_name)
+    except github3.exceptions.NotFoundError:
+        repo = None
 
     if repo is not None:
```
ethanhs: Nit: I think you can do try/except/else here too? Or did you find that it can return None?

dkess:
Thanks for the feedback, I've revised this PR to use a much cleaner style.

PR #129, opened 2019-04-30
Initial clean mypy run

ethanhs:
Because create can never be over engineered.

### DIFF
--- a/ircbot/plugin/stack_exchange.py
+++ b/ircbot/plugin/stack_exchange.py
@@ -1,9 +1,10 @@
 """Print information about Stack Exchange links."""
-import collections
 import functools
 import re
 import urllib.parse
 from datetime import datetime
+from typing import Dict
+from typing import NamedTuple
 
 import requests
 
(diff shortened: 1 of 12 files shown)


ircbot/plugin/stack_exchange.py
```diff
@@ -39,13 +43,12 @@ def _sites():
...
-            name=site['name'],
-        )
-        for site in resp.json()['items']
-    }
+    sites: Dict[str, Site] = {}
+    for site in resp.json()['items']:
+        domain = domain_from_url.match(site['site_url'])
+        if domain is not None:
+            sites[domain.group(1)] = Site(api_name=site['api_site_parameter'], name=site['name'])
+    return sites
```
dkess: Should we declare the return type of this function?

ethanhs: This PR is meant as a minimal clean run needed to get mypy passing and have a little bit of type checking. Most functions are still un-annotated. If you think this function in particular is worth annotating I can go ahead and annotate it, but I can also leave it for a later PR.

dkess: Oh it's np

PR #132, opened 2019-05-19
Add "wiitl" as a trigger for labstaff check

NotRyan:
\<fydai> petition to create a shorter command for that so it's easier for me to type while driving

### DIFF
--- a/ircbot/plugin/lab.py
+++ b/ircbot/plugin/lab.py
@@ -6,6 +6,7 @@ from ocflib.lab.stats import users_in_lab_count
 def register(bot):
     bot.listen(r'is ([a-z]+) in the lab', in_lab, require_mention=True)
     bot.listen(r"(who is|who's) in the lab", who_is_in_lab, require_mention=True)
+    bot.listen(r"(?i)wiitl", who_is_in_lab)
 
 
 def in_lab(bot, msg):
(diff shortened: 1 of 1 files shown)


cg505:
what is the `(?i)`?

redplanks:
Can't it be "!lab"?

cg505:
`!` is harder to type but that's probably also fine

ethanhs:
The original point is to make it easy to type on the phone. On most keyboards you have to change state to reach the `!` key so I would prefer the current implementation.

ethanhs:
@cg505 the `(?i)` is a regex flag to tell it to ignore case. Ctrl+F for `aiLmsux` in https://docs.python.org/3/library/re.html#regular-expression-syntax.

ircbot/plugin/lab.py
```diff
@@ -6,6 +6,7 @@
 def register(bot):
     bot.listen(r'is ([a-z]+) in the lab', in_lab, require_mention=True)
     bot.listen(r"(who is|who's) in the lab", who_is_in_lab, require_mention=True)
+    bot.listen(r"(?i)wiitl", who_is_in_lab)
```
ethanhs: please change this to single quotes so the lint passes. (The listen above uses double quotes do the apostrophe).

NotRyan:
> what is the `(?i)`?
makes the regex case insensitive

ethanhs:
Thank you @NotRyan! I'm sure fydai and others will appreciate this change.

PR #133, opened 2019-05-23
Rewrite of web server with Flask and slight refactoring of ircbot.py

fydai:
I moved as much plugin specific code from ircbot.py as possible, including a lot of celery-specific stuff and the timer code. I made a bot.threads attribute, and when any of the threads stops, the bot should go down with them (and then be restarted by Kubernetes or whatever).

I also rewrote the existing web server in Flask, so it should now be 1000x easier to add any any webhooks our hearts desire.
Some nice-to-include webhooks off the top of my head would be Prometheus, Github, RT, and Kanboard.


ircbot/ircbot.py
```diff
@@ -443,32 +376,17 @@ def main():
...
-        target=timer,
-        args=(bot,),
-        daemon=True,
-    )
-    timer_thread.start()
+    bot.threads.append(bot_thread)
 
     while True:
-        for thread in (bot_thread, celery_thread, timer_thread):
+        for thread in bot.threads:
```
dkess: I see this isn't a regression, since the code was like this before-- but is this wasting CPU cycles constantly checking? Is there a better way to do this?

fydai: Check every 5 seconds /s

ircbot/plugin/webserver.py
```diff
@@ -0,0 +1,48 @@
...
+
+def register(bot):
+    web_server_thread = threading.Thread(target=start_server, args=(bot,), daemon=True)
+    web_server_thread.start()
+    bot.threads.append(web_server_thread)
+
+
+@app.route('/', methods=['GET'])
+def route_base():
+    plugins: DefaultDict[ModuleType, Set[Listener]] = collections.defaultdict(set)
```
abizer: maybe build this outside of the function so it doesn't get rebuilt every time someone calls `ircbot.ocf.berkeley.edu`

fydai: That would require adding some global variables, which I don't really want to do. This is how it works currently and it seems to work fine (<10ms per hit).

ircbot/ircbot.py
fydai: In all seriousness I have no idea how this could be improved short of a rewrite with asyncio.

ircbot/plugin/webserver.py
abizer: you wouldn't be mutating a global variable, just instantiating it once and then reading it. it's basically a runtime constant, like reading in a config file.

ircbot/ircbot.py
keur: Here is an implementation. The `is_alive` method just checks a variable `is_stopped`, which is only mutated in two functions. If it gets triggered we can just trigger a callback, one of which could throw an exception. I don't actually endorse this method btw, I just got bored.

```python
import threading

class CallbackThread(threading.Thread):
    def __init__(self, group=None, target=None, name=None,
            args=(), kwargs=None, *, callback=None, daemon=None):
        super().__init__(group=group, target=target, name=name, args=args, kwargs=kwargs, daemon=daemon)
        self._callback = callback

    def _callback_if_dead():
        if self._is_stopped:
            self._callback()

    def _stop(self):
        super()._stop()
        self._callback()

    def _reset_internal_locks(self, is_alive):
        super()._reset_internal_locks(is_alive)
        self._callback_if_dead()
```

fydai: In a newer commit, I added a wrapper around these threads, if they exit then it terminates the entire program.

ircbot/plugin/webserver.py
fydai: Changed as requested.

PR #135, opened 2019-05-28
make `wiitl` work like summoning piiinnngg

MrMinos:
To facilitate easier typing experience on mobile, with this PR, one can doubly type a character while summoning create with command for `who is in the lab` and create will understand it.


MrMinos:
For comments that I could anticipate:

Is it necessary? No.  : D

MrMinos:
ty dkess!

PR #136, opened 2019-05-29
Add kubernetes and remove marathon

fydai:


Jenkinsfile
```diff
@@ -1,6 +1,6 @@
 servicePipeline(
     upstreamProjects: ['ocflib/master', 'dockers/master'],
-    deployTargets: ['ircbot'],
+    deployTargets: [],
```
dkess: You can just delete this line

fydai: I don't think you can, Jenkins failed when I tried it before.

PR #139, opened 2019-06-17
weather: switch to openweathermap API

dkess:


dkess:
This PR also slightly changes the display-- we get rid of forecasts (which were broken anyways) and adds description, humidity, wind, and UV index.

dkess:
If we end up being unhappy with OpenWeatherMap's built-in city name detection, we can use LocationIQ (which also has a free API) to get coordinates and then plug those into OpenWeatherMap.

dkess:
Sample output:

```
11:55 <@dkessler> create-dkessler: weather
11:55 <create-dkessler> Berkeley: 68.43°F ☀ few clouds; wind: 8.05m/s 260° (W); humidity: 56; UV index: 9.06
11:55 <@dkessler> create-dkessler: weather manhattan
11:55 <create-dkessler> Manhattan: 86.14°F ☢ clear sky; wind: 12.75m/s 210° (SSW); humidity: 55; UV index: 9.18
```

When viewed in a proper IRC client there are also colors and boldtext which makes this easier to read at a glance.

PR #140, opened 2019-06-25
weather: handle case with unspecified wind dir

dkess:


ircbot/plugin/weather.py
```diff
@@ -136,11 +136,19 @@ def get_summary(api_key, location, unit='f'):
...
     ico = icon(temp, unit=unit)
+
     desc = j['weather'][0]['description']
 
     windspeed = j['wind']['speed']
-    winddeg = j['wind']['deg']
-    wind_text = f'{windspeed}m/s {winddeg}° ({deg_to_compass(winddeg)})'
+
+    winddeg_text = ''
+    try:
```
cg505: cant we just do `if 'deg' in j['wind']:` instead of this try/except?

ircbot/plugin/weather.py
```diff
@@ -136,11 +136,19 @@ def get_summary(api_key, location, unit='f'):
...
+        winddeg = j['wind']['deg']
```
cg505: i know you didn't name this variable `j` but i'm not a fan

ircbot/plugin/weather.py
```diff
@@ -136,11 +136,19 @@ def get_summary(api_key, location, unit='f'):
...
+        winddeg_text = f'{winddeg}° ({deg_to_compass(winddeg)})'
+    except KeyError:
+        pass
+
+    wind_text = f'{windspeed}m/s {winddeg_text}'.strip()
```
cg505: instead of using `.strip()` just put the space in `winddeg_text` imo

dkess: Not sure what you mean here. The point is the remove the trailing space in case `winddeg_text` is empty.

ircbot/plugin/weather.py
dkess: This can be changed in a later commit

ircbot/plugin/weather.py
Baisang: or `j['wind'].get('deg', <some default>)`

dkess: A default won't work here because then the format statement afterwards won't make sense. I'll go with @cg505's solution.

ircbot/plugin/weather.py
cg505: right, so instead of using `.strip()`, put the trailing space in `winddeg_text`. So we either have something like `winddeg_text = ' 0° (N)'` (note leading spae) or `winddeg_text = ''`

dkess: Ohhhh I see, sgtm

Issue #114, opened 2019-03-18
add webserver smoketest


keur:
#113 caused `jinja` to hard fail since the `show_topic` method didn't have a docstring. This is especially annoying because Marathon uses the webserver to healthcheck the bot—so a broken webserver leaves ircbot in a deploy loop. We can't reasonably expect people to remember to docstring methods in the plugin class. We should have `make test` run a smoke test that ensures the webserver 200s

PR #143, opened 2019-07-23
Add !still command

asottile:
Based on my twitch irc bot's command: https://github.com/anthonywritescode/twitch-chat-bot/blob/e8f3093fb4acb31bd89942c16110f5a8e78f6ea4/bot.py#L160-L166

### DIFF
--- a/ircbot/plugin/still.py
+++ b/ircbot/plugin/still.py
@@ -1,0 +1,13 @@
+"""Really? it's {current_year}."""
+import datetime
+import random
+
+
+def register(bot):
+    bot.listen(r'^!still (.+)', still)
+
+
+def still(bot, msg):
+    year = datetime.date.today().year
+    lol = random.choice(['LOL', 'LOLW', 'LMAO', 'NUUU'])
+    msg.respond(f'{msg.match[1]}, in {year} - {lol}!')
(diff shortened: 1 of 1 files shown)


ircbot/plugin/still.py
```diff
@@ -0,0 +1,13 @@
+"""Really? it's {current_year}."""
+import datetime
+import random
+
+
+def register(bot):
+    bot.listen(r'^!still (.+)', still)
+
+
+def still(bot, msg):
```
cg505: Please add a docstring to this function.

asottile: why, the module and function is self explanatory

cg505: It will break the help page if you don't have one.

asottile: I'll add a docstring when there's an automated test having me add a docstring

cg505: #114 has been open for a bit but it's relatively low priority so no one has gotten to it. Unfortunately this PR can't be merged without the docstring because it will break the deploy. I didn't mean to come across as a jerk but this will break the bot if merged as-is.

asottile: It won't break the deploy, None values are skipped by jinja

PR #144, opened 2019-07-23
Make help non-nullable

asottile:


ircbot/ircbot.py
```diff
@@ -64,11 +63,13 @@ class Listener(NamedTuple):
...
     @property
-    def help(self) -> Optional[str]:
+    def help(self) -> str:
         if self.help_text:
             return self.help_text
-        else:
+        elif self.fn.__doc__:
             return self.fn.__doc__
+        else:
+            return self.pattern.pattern
```
cg505: imo this doesn't make much sense since the only place we use this is, it's displayed right next to the pattern anyway. Maybe we should say "No help text" or something instead.

PR #146, opened 2019-08-14
add "has a small truck" to !sux

ja5087:
ask and ye shall receive


cg505:
smh merging over tests

ja5087:
chad move

PR #148, opened 2019-08-31
google: add !yt command for youtube search

dkess:
This adds a layer of abstraction for making general Google searches, and adds a youtube function that does the same thing as regular Google but adds site:youtube.com to the query.


kpengboy:
Was the alternative of using the YouTube API considered?

dkess:
I considered it but this is clearly way easier

PR #153, opened 2019-09-22
don't cast F2C conversion to int

Aesthetic:
prevents rounding errors that show the wrong emoji when displaying temp as Celsius


kpengboy:
I would prefer the methodology of PR #154. (In any case though, thanks for the contribution.)

PR #154, opened 2019-09-22
Weather plugin fixes

steven676:
Fixes a couple of issues with display in the weather plugin.

### DIFF
--- a/ircbot/plugin/weather.py
+++ b/ircbot/plugin/weather.py
@@ -144,9 +144,13 @@ def get_summary(api_key, location, unit='f'):
     winddeg_text = ''
     if 'deg' in j['wind']:
         winddeg = j['wind']['deg']
-        winddeg_text = f' {winddeg}° ({deg_to_compass(winddeg)})'
+        winddeg_text = f'{winddeg}° ({deg_to_compass(winddeg)})'
 
-    wind_text = f'{windspeed}m/s{winddeg_text}'
+    windunit_text = 'm/s'
+    if unit == 'f':
+        windunit_text = 'mph'
+
+    wind_text = f'{windspeed} {windunit_text} {winddeg_text}'
 
     humidity = j['main']['humidity']
 
@@ -155,6 +159,6 @@ def get_summary(api_key, location, unit='f'):
     return '; '.join([
         f'{name}: {bold(color(temp, unit=unit))} {ico} {bold(desc)}',
         f'wind: {bold(wind_text)}',
-        f'humidity: {bold(humidity)}',
+        f'humidity: {bold(humidity)}%',
         f'UV index: {bold(uv_index)}',
     ])
--- a/ircbot/plugin/weather.py
+++ b/ircbot/plugin/weather.py
@@ -20,8 +20,8 @@ def weather(bot, msg):
         msg.respond(f'idk where {where} is')
 
 
-def f2c(temp):
-    return int((temp - 32) * 5 / 9)
+def c2f(temp):
+    return temp * (9 / 5) + 32
 
 
 def deg_to_compass(deg):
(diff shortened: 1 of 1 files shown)


ircbot/plugin/weather.py
```diff
@@ -58,11 +57,14 @@ def deg_to_compass(deg):
 
 
 def icon(temp, unit='f'):
```
kpengboy: Nit: do we ever call this function (or `color`) without a `unit` argument? If not, then there is not much point in making it optional.

ircbot/plugin/weather.py
```diff
@@ -146,7 +151,11 @@ def get_summary(api_key, location, unit='f'):
         winddeg = j['wind']['deg']
         winddeg_text = f' {winddeg}° ({deg_to_compass(winddeg)})'
 
-    wind_text = f'{windspeed}m/s{winddeg_text}'
+    windunit_text = 'm/s'
+    if unit == 'f':
+        windunit_text = 'mph'
+
+    wind_text = f'{windspeed}{windunit_text}{winddeg_text}'
```
kpengboy: Nit: would put space between number and unit

ircbot/plugin/weather.py
```diff
@@ -73,14 +75,17 @@ def bold(text):
 
 
 def color(temp, text=None, unit='f'):
+    if unit == 'c':
+        temp_f = c2f(temp)
+    else:
+        temp_f = temp
```
kpengboy: Nit: may be more compact using a ternary condition?

`temp_f = c2f(temp) if unit == 'c' else temp`

ircbot/plugin/weather.py
```diff
@@ -20,9 +20,8 @@ def weather(bot, msg):
         msg.respond(f'idk where {where} is')
 
 
-def f2c(temp):
-    return int((temp - 32) * 5 / 9)
-
+def c2f(temp):
+    return temp * 9/5 + 32
```
kpengboy: Make sure to insert 1 more blank line so this file passes style check (and therefore doesn't break the build)

ircbot/plugin/weather.py
```diff
@@ -58,11 +57,14 @@ def deg_to_compass(deg):
 
 
 def icon(temp, unit='f'):
+    if unit == 'c':
+        temp = c2f(temp)
+
     # it doesn't get cold in the bay area, anything less than 60°F is a snowman
-    if temp < (60 if unit == 'f' else f2c(60)):
+    if temp < 60:
```
Aesthetic: does this still check correctly? it looks to be F only

ircbot/plugin/weather.py
steven676: Nope -- will make the argument mandatory, as suggested.

ircbot/plugin/weather.py
steven676: I was following the original behavior here -- we can certainly have a space, though.

ircbot/plugin/weather.py
steven676: There's a c2f() conversion for temp just above.

steven676:
Okay, now with the requested changes (plus another style fix to shut the checks up) ...

ircbot/plugin/weather.py
Aesthetic: that's the issue, it still checks against the hardcoded 60 F value

ircbot/plugin/weather.py
```diff
@@ -72,15 +75,14 @@ def bold(text):
     return f'\x02{text}\x0F'
 
 
-def color(temp, text=None, unit='f'):
-    if text is None:
-        text = '{}°{}'.format(temp, unit.capitalize())
+def color(temp, unit):
+    temp_f = c2f(temp) if unit == 'c' else temp
+
+    text = '{}°{}'.format(temp, unit.capitalize())
```
ethanhs: Optional nit: this can be changed into an f string! Something like 

`text = f'{temp}°{unit.capitalize()}'`

steven676: That's a pretty big win for readability, let's do that.

PR #158, opened 2019-11-05
webserver: add webhook for prometheus alerts

dkess:
This is pretty much a copy of the code from #76.

We'll start off with the #rebuild-spam channel for now, but if this works out well we can extend this to use custom channels (configured by Prometheus) and then utilize the #opstaff-alerts channel.

It looks like this:

```
<create> OK TestAlert: This is what a Prometheus alert looks like
```

In IRC, `OK` or `FIRING` show up as green and red respectively, and the part before the colon is bold.

### DIFF
--- a/ircbot/plugin/webserver.py
+++ b/ircbot/plugin/webserver.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING
 
 from flask import Flask
 from flask import render_template
+from flask import request
 
 if TYPE_CHECKING:
     from ircbot.ircbot import Listener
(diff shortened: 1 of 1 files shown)


dkess:
Oh actually can we use f-strings now? 🤔

ircbot/plugin/webserver.py
```diff
@@ -52,6 +53,25 @@ def route_macros():
...
+    for alert in body_json['alerts']:
+        if alert['status'] == 'resolved':
+            status = '\x02\x0303OK\x0F'
+        else:
+            status = '\x02\x0304FIRING\x0F'
+
+        alert = '{status} \x02{alertname}\x0F: {summary}'.format(
+            status=status,
+            alertname=alert['labels']['alertname'],
+            summary=alert['annotations']['summary'],
```
keur: I would be extra safe and do a check here that these exist in the dictionary. The annotations section is entirely optional (if you don't believe me, you can verify this with `promtool`). Right now all the alerts we have a summary, but we don't have documentation for OCF alert standards, so some could eventually slip in without it.

ircbot/plugin/webserver.py
```diff
@@ -52,6 +53,25 @@ def route_macros():
     )
 
 
+@app.route('/hook/prometheus', methods=['POST'])
+def route_prometheus():
```
keur: what is stopping me from sending a fake alert

dkess: Nothing, but if it becomes a problem we can add an HTTP password check.

PR #159, opened 2019-11-05
Add overly-safe synchronization

fydai:
To be safe, we synchronize pretty much everything (responding to messages, timer-triggered "bot.say", etc). This should make it safe to perform bot methods (such as "say") from multiple threads (in particular, the webserver and the timer thread).

### DIFF
--- a/ircbot/ircbot.py
+++ b/ircbot/ircbot.py
@@ -54,6 +54,15 @@ NUM_RECENT_MESSAGES = 10
 MAX_CLIENT_MSG = 435
 
 
+def synchronize(method):
+    """Decorator to wrap a method in a lock-acquiring context manager"""
+    @functools.wraps(method)
+    def new_method(self, *args, **kwargs):
+        with self.lock:
+            return method(self, *args, **kwargs)
+    return new_method
+
+
 class Listener(NamedTuple):
     pattern: Pattern
     fn: FunctionType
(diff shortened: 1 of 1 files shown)


fydai:
Sure, we should only lock the shared resources, but I contend that we should just lock everything. It's not as elegant, there will be a completely negligible performance impact, but it will let me feel safe merging useful features like https://github.com/ocf/ircbot/pull/158. If you would like to clearly identify all shared resources between the threads, and lock solely those resources, that would be much appreciated. 

I also disagree that it's "brittle", because it's overly safe. There only bad thing we can do is by causing deadlock by running a synchronized function endlessly (which we don't do). Python has its GIL, which enforces that only one thread can do anything at a time. This code just merely ensures that the only thread switching can occur when the bot is done processing its current thing.

I do agree it's worth exploring async, but that would require almost a complete rewrite of ircbot, something which is very much not a priority.

keur:
On Mon Nov 4, 2019 at 10:47 PM fydai wrote:
> Sure, we should only lock the shared resources, but I contend that we
>should just lock everything.

You aren't locking everything. If you want to do that, put a lock on the
bot object.

>It's not as elegant, there will be a completely negligible performance
i>impact, but it will let me feel safe merging useful features
>like https://github.com/ocf/ircbot/pull/158.

We already "feel safe" using bot.say in the timer thread. Introducing
half-baked locks is not going to provide long term safety.

>If you would like to clearly identify all shared resources between the
>threads, and lock solely those resources, that would be much
>appreciated.

No. If you want thread safety you can do it correctly instead of
implying that I ought to do work for you.

> I also disagree that it's "brittle", because it's overly safe.

The entire bot object is being shared between threads. It's brittle
because if new attributes of the bot are read/written to in threads you
have the same problem. Further, you have indicated that you have not
identified all cases of shared resource usage, so this might not even
cover all the existing cases. This patch is only overly safe in that you
are locking methods like on_pubmsg that will only ever be called in one
thread.

PR #162, opened 2019-11-17
DSA error handling: return handled error string in a list of lines

MrMinos:


MrMinos:
will come back to address this in the near future.

briefly for (3), do you mean when we actually fetch valid DSAs? or the error text from DSA?

dkess:
In (3) i was talking about the text of valid DSAs

PR #163, opened 2019-11-17
modify timer to wait longer periods if a fetch fails, up till max of 30m

MrMinos:


MrMinos:
note: using the lower case variable name, as it is not a constant anymore

PR #164, opened 2019-11-18
refactor DSA backoff logic and disable temporarily

cg505:


ircbot/plugin/timer.py
```diff
@@ -52,6 +56,6 @@ def timer(bot):
                     traceback=format_exc(),
                 ),
             )
-            dsa_freq = dsa_freq + 5 if dsa_freq < 30 else dsa_freq
+            dsa_freq = dsa_freq + DSA_FREQ_BACKOFF if dsa_freq < DSA_FREQ_MAX else dsa_freq
```
dkess: If I'm understanding this correctly, this won't converge to `DSA_FREQ_MAX` at the end of the timer. I'd make this the `max` of the regular `dsa_freq` increment and `DSA_FREQ_MAX`.

PR #165, opened 2019-11-19
Revert "disable DSA checking while they're throwing 403s"

cg505:
This reverts commit 712f6350d220be36ebc8d8cf384624123ee5f4dd.


MrMinos:
I'd bet a dollar to nuts fund that this will come back : P

PR #166, opened 2019-12-15
Replace flask dev server with python wsgi server

fydai:


requirements.txt
```diff
@@ -1,60 +1,60 @@
...
-ldap3==2.6
-markovify==0.7.1
+kombu==4.6.7
+ldap3==2.6.1
+markovify==0.7.2
 MarkupSafe==1.1.1
-more-itertools==7.0.0
-ocflib
+more-itertools==8.0.2
+ocflib==2019.11.12.16.30
```
dkess: remove version number

ircbot/plugin/webserver.py
```diff
@@ -53,6 +65,9 @@ def route_macros():
 
 
 def start_server(bot):
-    port = os.getenv('HTTP_PORT', 8888)
-    app.bot = bot
-    app.run(host='0.0.0.0', port=int(port))
+    global create_bot
```
dkess: I'm pretty sure using global variables and shared memory in this way is undefined behavior in WSGI's process model, which is why I avoided doing something similar. If this really works reliably then I'm probably OK with it, but I'd make sure to test this extra carefully.

cg505: You're right, which is why we can't use gunicorn for instance. I believe cheroot was picked specifically because it's implemented in Python so this kind of hack will work. Otherwise things will need to be drastically refactored in a painful way. But yeah, this is not ideal.

PR #167, opened 2019-12-15
Add a lock to the bot's say function

fydai:
We are currently accessing the say function from multiple threads (celery, web server, and standard server) all use it. Adding a lock would remove any possible synchronization issues from using multiple threads.

### DIFF
--- a/ircbot/ircbot.py
+++ b/ircbot/ircbot.py
@@ -135,6 +135,9 @@ class CreateBot(irc.bot.SingleServerIRCBot):
         self.plugins: Dict[str, ModuleType] = {}
         self.extra_channels: Set[str] = set()  # plugins can add stuff here
 
+        # As we use threads, we should ensure that we use them safely
+        self.lock = threading.RLock()
+
         # Register plugins before joining the server.
         self.register_plugins()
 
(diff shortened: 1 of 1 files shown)


ircbot/ircbot.py
```diff
@@ -135,6 +135,9 @@ def __init__(
         self.plugins: Dict[str, ModuleType] = {}
         self.extra_channels: Set[str] = set()  # plugins can add stuff here
 
+        # As we use threads, we should ensure that we use them safely
+        self.lock = threading.RLock()
```
dkess: Why did you use an `RLock` here? It doesn't look to me like this lock will ever be taken twice in the same thread.

cg505:
We should look other interactions that use the IRC connection, like changing the topic.

dkess:
By all means look into it, but if we let perfect be the enemy of the good here the IRC bot will remain stagnant forever. I would argue that any effort spent on getting the synchronization exactly right is better directed towards refactoring the bot entirely.

ircbot/ircbot.py
```diff
@@ -340,17 +343,19 @@ def plusone(m):
...
-            self.connection.privmsg(channel, message)
+        # This is writing to a socket, it should be synchronized
+        with self.lock:
+            # Find the length of the full message
+            msg_len = len(f'PRIVMSG {channel} :{message}\r\n'.encode('utf-8'))
+
+            # The message must be split up if over the length limit
+            if msg_len > MAX_CLIENT_MSG:
+                messages = split_utf8(message.encode('utf-8'), MAX_CLIENT_MSG)
+
```
tzhu-yelp: This math portion doesn't need the lock.

tzhu-yelp:
> We should lock other interactions that use the IRC connection, like changing the topic.

This needs a lock because multiple IRC messages are being sent, that's not the case when changing the topic.

nattofriends:
> We should lock other interactions that use the IRC connection, like changing the topic.

This needs a lock because multiple messages are being sent. Changing the topic doesn't create multiple messages.

ircbot/ircbot.py
nattofriends: This math part doesn't need the lock. Only sending multiple messages needs the lock.

cg505:
The lock should be held whenever the socket needs to be used (that is, whenever we interact with the IRC server) because sockets in Python are not thread safe.

ircbot/ircbot.py
gnowxilef: +1

PR #170, opened 2020-03-31
Add `update` keyword as alias for `replace`

encadyma:
* Updated shorturls and macros to use alias
* Alias could be edited to use a proxy function that can appear differently on ircbot.ocf.berkeley.edu


encadyma:
That sounds like a much cleaner solution @jvperrin ! Thanks for the suggestion, I've updated with the change.

Issue #172, opened 2020-04-16
Log directly to stdout


cg505:
Now that we are using kubernetes logging, the container should not use stdin2syslog as in https://github.com/ocf/ircbot/blob/4ae53dd1790d09e65abc69453c37d838c3325a8f/services/ircbot/log/run#L3
We can probably delete the whole services directory, and simply modify the Dockerfile to run the command https://github.com/ocf/ircbot/blob/4ae53dd1790d09e65abc69453c37d838c3325a8f/services/ircbot/run#L6 instead.

This should allow us to see logs by running something like `kubectl -n app-ircbot get logs deployment/deployment` instead of having to connect to democracy to see logs.

PR #173, opened 2020-05-25
Logs go directly to stdout

pxhanus:
Fixes issue #172


dkess:
You can remove `runit` from the Dockerfile too

pxhanus:
> You can remove `runit` from the Dockerfile too

Addressed

PR #174, opened 2020-05-29
xkcd

nikhiljha:
I wrote 20 short programs in Python yesterday.  It was wonderful.  Perl, I'm leaving you.


ircbot/plugin/xkcd.py
```diff
@@ -0,0 +1,21 @@
...
+def register(bot):
+    bot.listen(REGEX.pattern, show_comic)
+
+
+def show_comic(_, msg):
+    """Show XKCD comic details."""
+    for number in REGEX.findall(msg.text):
+        try:
+            comic = xkcd.Comic(int(number))
+            reply = f"XKCD#{int(number)} | '{comic.getTitle()}' | {comic.getImageLink()}"
```
dkess, line 18: Link to the comic instead of the image IMO

cg505: linking the image directly will cause it to expand in matrix/slack. this message might be responding to a link anyway

jvperrin: You definitely lose the hover/alt text when just having the image, which detracts from the comic. That could be posted in a second line or something but I do think it needs to be either included or the full comic linked to.

ircbot/plugin/xkcd.py
```diff
@@ -0,0 +1,21 @@
...
+            msg.respond(str(reply))
```
ja5087, line 19: fstrings are already strings?

ircbot/plugin/xkcd.py
```diff
@@ -0,0 +1,21 @@
...
+        except AssertionError:
```
ja5087, line 20: Where is this meant to catch?

ircbot/plugin/xkcd.py
nikhiljha: I was considering linking to explainxkcd too, that has everything + an explanation if you need it :P

ja5087: I think just post the image and alt text. The point is to present the comic inline within the conversation, not so we have to click another link.

cg505: yeah, alt text is important

ircbot/plugin/xkcd.py
nikhiljha: oops, was supposed to check `HTTPError` in case it's not a valid comic (and therefore 404s)

nikhiljha:
Ok, tested and should work now 👍

PR #189, opened 2021-01-24
Add notify plugin

cg505:
This adds a mechanic for adding "notification groups" that you can use to quickly ping a subgroup of people (sort of like Slack teams, which are a paid feature). Include something like `!!gm` or `!!finance-comm` to expand to ping each individual member in those notification groups.

Notification groups are comprised of owners and members. Both are pinged during normal expansion. Additionally, owners have permissions to add other users or owners.

Anyone can join any group by using `addme` and leave by using `removeme` (as a member, not an owner).

Finally groups can be nested by adding `!!groupname` to another group. This also allows you to create aliases, by creating a group where the only member is the canonical group.

### DIFF
--- a/ircbot/plugin/notify.py
+++ b/ircbot/plugin/notify.py
@@ -1,0 +1,491 @@
+"""Ping multiple users at once in notification groups."""
+from functools import wraps
+from typing import Optional
+from typing import Set
+
+from ircbot import db
+
+MAX_DEPTH = 100
+
+
+class NotifyExpansionLoopError(Exception):
+    pass
+
+
+def register(bot):
+    bot.listen(r'!!([^ ]+)', notify)
+    bot.listen(r'^!notify show ([^ ]+)$', show)
+    bot.listen(r'^!notify showdumb ([^ ]+)$', showdumb)
+    bot.listen(r'^!notify create ([^ ]+)$', create)
+    bot.listen(r'^!notify delete ([^ ]+)$', delete)
+    bot.listen(r'^!notify list$', list_groups)
+    bot.listen(r'^!notify addme ([^ ]+)$', addme)
+    bot.listen(r'^!notify (?:remove|delete)me ([^ ]+)$', removeme)
+    bot.listen(r'^!notify addmembers? ([^ ]+) (.+)$', addmembers)
+    bot.listen(r'^!notify addowners? ([^ ]+) (.+)$', addowners)
+    bot.listen(r'^!notify (?:remove|delete) ([^ ]+) (.+)$', remove)
+    bot.listen(r'^!notify clear ([^ ]+)$', clear)
+    bot.listen(r'^!notify add ', addhelp)
+
+
+def handle_loop_error(fn):
+    """Decorate fn to manually catch expansion errors, so they don't send emails."""
+    @wraps(fn)
+    def handled_fn(bot, msg):
+        try:
+            return fn(bot, msg)
+        except NotifyExpansionLoopError as e:
+            msg.respond(str(e))
+    return handled_fn
+
+
+def get_group(cursor, slug: str):
+    cursor.execute(
+        'SELECT * FROM notify WHERE slug = %s',
+        (slug,),
+    )
+    return cursor.fetchone()
+
+
+def expand(cursor, targets: str, depth: int = 0) -> Set[str]:
+    """Perform expansion on the targets string."""
+    result = set()
+    target_arr = targets.split(' ')
+    for target in target_arr:
+        if target[:2] != '!!':
+            result.add(target)
+        else:
+            subtargets = get_all_targets(cursor, target[2:], depth + 1)
+            if subtargets is not None:
+                result |= subtargets
+    return result
+
+
+def get_all_targets(cursor, slug: str, depth: int = 0) -> Optional[Set[str]]:
+    """Expand the full unique list of owners and members of slug."""
+    if depth > MAX_DEPTH:
+        raise NotifyExpansionLoopError(f'Hit MAX_DEPTH while expanding !!{slug}. Do you have a loop?')
+
+    subtarget = get_group(cursor, slug)
+    if subtarget is None:
+        return None
+
+    result = set()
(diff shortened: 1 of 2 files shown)


encadyma (approved):
pretty excited for this

cg505:
okay in retrospect maybe it would be better if the owners and members columns were both non-null

ircbot/plugin/notify.py
```diff
@@ -0,0 +1,478 @@
...
+def get_group(cursor, slug: str):
+    cursor.execute(
+        'SELECT * FROM notify WHERE slug = %s',
+        (slug,),
+    )
+    return cursor.fetchone()
+
+
+def expand(cursor, targets: str, depth: int = 0) -> Set[str]:
+    """Performance expansion on the targets string."""
```
ethanhs, line 49:
```suggestion
    """Perform expansion on the targets string."""
```

Typo I think?

ircbot/plugin/notify.py
```diff
@@ -0,0 +1,478 @@
...
+    bot.listen(r'^!notify showdumb ([^ ]+)$', showdumb)
+    bot.listen(r'^!notify create ([^ ]+)$', create)
+    bot.listen(r'^!notify delete ([^ ]+)$', delete)
+    bot.listen(r'^!notify addme ([^ ]+)$', addme)
+    bot.listen(r'^!notify (?:remove|delete)me ([^ ]+)$', removeme)
+    bot.listen(r'^!notify addmembers? ([^ ]+) (.+)$', addmembers)
+    bot.listen(r'^!notify addowners? ([^ ]+) (.+)$', addowners)
+    bot.listen(r'^!notify (?:remove|delete) ([^ ]+) (.+)$', remove)
+    bot.listen(r'^!notify clear ([^ ]+)$', clear)
+    bot.listen(r'^!notify add ', addhelp)
```
ethanhs, line 26: Are you trying to correct people using e.g. `!notify add myname`? If so shouldn't you capture everything after `add`?

ethanhs (commented):
This looks really handy, thank you for adding this. I just have a question and a minor typo correction (I think).

ircbot/plugin/notify.py
cg505: The idea is that if you do `!notify add <i dont care what you put here>`, it's unclear if you mean `create`, `addme`, `addmembers` or `addowners`. So we just tell you to pick one of those instead. Since we don't care about anything after the "add" I don't bother including it in the regex. You'll see that there's no `$` at the end of the regex.

ircbot/plugin/notify.py
```diff
@@ -0,0 +1,478 @@
...
+    except NotifyExpansionLoopError as e:
+        respond(f'During owner check: {e} Using rudimentary check instead.', ping=False)
+        return nick in group['owners'].split(' ')
+
+
+@handle_loop_error
+def notify(bot, msg):
+    """Notify all targets (owners and members) in the tagged notificaton group."""
+    slug = msg.match.group(1)
+    with db.cursor(password=bot.mysql_password) as c:
```
jvperrin, line 135: Maybe make this into a decorator or something shared among functions since it's used in a bunch of them?

ircbot/plugin/notify.py
```diff
@@ -0,0 +1,478 @@
...
+        try:
+            return fn(bot, msg)
+        except NotifyExpansionLoopError as e:
+            msg.respond(str(e))
+    return handled_fn
+
+
+def get_group(cursor, slug: str):
+    cursor.execute(
+        'SELECT * FROM notify WHERE slug = %s',
```
jvperrin, line 42: Is there a schema for this anywhere? I've seen that for some other tables in ocflib or something (https://github.com/ocf/ocflib/blob/master/ocflib/misc/ocfshorturls.sql for instance), but not sure where a schema for this would even go since it feels a bit strange just dumping it into ircbot...

jvperrin (commented):
I think one useful command would be to list what groups are out there.

I'd also be more in favor of keeping all groups flat and not allowing nesting, since that seems like it would make so much of this simpler and easy to reason about. I do like that they have owners though, as that seems like it'll make it easier to update and keep current than relying on only IRC opers.

ircbot/plugin/notify.py
cg505: You can go in the database and look at the schema...

Is the schema for the quotes or macros database stored anywhere? If so, I can add it there. Mostly I was under the impression that these tables were pretty ad-hoc, which I don't think is a massive problem since their schemas are very simple and can be easily inferred if necessary, and the database is backed up anyway.

ircbot/plugin/notify.py
cg505: I don't think it's worth it. It wouldn't end up being shorter (since the shared code is only this one line). I think that the potential for bugs caused by updating some but not all is low, since the code is very simple and also literally identical everywhere.

cg505:
> I think one useful command would be to list what groups are out there.
Sure, should be easy enough.

> I'd also be more in favor of keeping all groups flat and not allowing nesting, since that seems like it would make so much of this simpler and easy to reason about.
Yeah this is a tough tradeoff. Personally I did the nested groups because it felt like it would be useful. If we don't have nesting, we will probably want to add some other way to do aliases. Also, it was more fun/interesting to implement the nested version. That said, if nesting ends up being rarely used and not that helpful, we probably should remove it.

ethanhs (approved):

ircbot/plugin/notify.py
jvperrin: Sure, I guess I was thinking the indentation would benefit but it's a pretty minor thing either way

ircbot/plugin/notify.py
jvperrin: That's true, I think the schema just makes it easier to look at for people who don't have ocfroot access yet. Good point about it already being backed up though, that does mean it's not really useful for any recovery.

jvperrin (approved):

PR #191, opened 2021-01-26
Update requirements

ethanhs:
*This pull request was made automatically by a human*


ethanhs:
This is actually stalled on updating ocflib, so I'll do that now.

ircbot/db.py
```diff
@@ -5,16 +5,17 @@
...
+        host='mysql.ocf.berkeley.edu',
+        cursorclass=pymysql.cursors.DictCursor,
+        charset='utf8mb4',
+        autocommit=True,
+    )
+    try:
+        with conn.cursor() as cursor:
             yield cursor
+    finally:
+        conn.close()
```
cg505, line 21: what is the point of using a `with` expression if you still have to do this?

ircbot/ircbot.py
```diff
@@ -146,7 +147,10 @@ def __init__(
 
     def register_plugins(self):
         for importer, mod_name, _ in pkgutil.iter_modules(['ircbot/plugin']):
-            mod = importer.find_module(mod_name).load_module(mod_name)
+            assert isinstance(importer, PathEntryFinder)
+            loader = importer.find_module(mod_name)
+            assert loader is not None
```
cg505, line 152: why are these `assert`s now necessary?

cg505 (approved):
just a couple questions

ircbot/db.py
ethanhs: The with is used to manage the context of the cursor, the connect/close doesn't have a contextmanager :/

ircbot/ircbot.py
ethanhs: These are assertions to please the type checker which can't tell that `importer` is always a `PathEntryFinder` and unless importing is broken the loader shouldn't be `None`.

PR #200, opened 2021-04-18
Add AQI output to weather

ja5087:
The openweathermap API only returns a score from 1-5 so I calculate the AQI using the EPA formula.

image

### DIFF
--- a/ircbot/plugin/weather.py
+++ b/ircbot/plugin/weather.py
@@ -110,7 +110,7 @@ def color(temp, unit):
 
 
 def calculate_aqi(conc):
-    """Calculate US EPA AQI score from PM 2.5 concentration in ug/m3 using
+    """Calculate US EPA AQI score from PM 2.5 concentration in µg/m3 using
        using the equation at https://forum.airnowtech.org/t/the-aqi-equation/169"""
 
     conc_lo_cutoffs = [0.0, 12.1, 35.5, 55.5, 150.5, 250.5]
--- a/ircbot/plugin/weather.py
+++ b/ircbot/plugin/weather.py
@@ -119,7 +119,7 @@ def calculate_aqi(conc):
     aqi_hi_cutoffs = [50, 100, 150, 200, 300, 500]
 
     conc = round(conc, 1)
-    t_idx = [i for i in range(len(conc_lo_cutoffs) - 1) if conc_lo_cutoffs[i] <= conc][-1]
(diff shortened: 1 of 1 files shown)


encadyma (approved):

ircbot/plugin/weather.py
```diff
@@ -109,6 +109,26 @@ def color(temp, unit):
...
 
 
+def calculate_aqi(conc):
+    """Calculate US EPA AQI score from PM 2.5 concentration in µg/m3 using
+       using the equation at https://forum.airnowtech.org/t/the-aqi-equation/169"""
+
+    conc_lo_cutoffs = [0.0, 12.1, 35.5, 55.5, 150.5, 250.5]
+    conc_hi_cutoffs = [12.0, 35.4, 55.4, 150.4, 250.4, 500.4]
+    aqi_lo_cutoffs = [0, 51, 101, 151, 201, 301]
+    aqi_hi_cutoffs = [50, 100, 150, 200, 300, 500]
```
cg505, line 119: this looks so incredibly stupid but i understand that it's directly from their algorithm.

ja5087:
The values aren't consistent across multiple websites, but I checked the output from the API and calculated it https://www.airnow.gov/aqi/aqi-calculator-concentration/ and the results are correct

cg505 (approved):

ircbot/plugin/weather.py
ja5087: I see your point, but I'm not sure what a clearer way to implement a piecewise linear function would be

nattofriends: Well honestly conc_lo_cutoffs are just a fragment of `For all AQI calculations, the calculated average concentrations are truncated to .001 ppm (1 ppb) for O3, 0.1 μg/m3 for PM2.5, and 1 μg/m3 for PM10.`, so I think it might be good enough to do something easier, if these variables offend you so:

```python
import itertools

AQI_BUCKETS = (
  (0.0, 50),
  (12.1, 50),
  (35.5, 50),
  (55.5, 50),
  (150.5, 100),
  (250.5, 100),
)

def calculate_aqi(conc):
  aqi = 0

  a, b = itertools.tee(AQI_BUCKETS)
  next(b)
  
  for (conc_lo, bucket), (conc_lo_next, _) in zip(a, b):
    print(conc_lo, bucket, conc_lo_next)
    if conc > conc_lo_next:
      aqi += bucket
    else:
      aqi += bucket * ((conc - conc_lo) / (conc_lo_next - conc_lo))
      return round(aqi)
```

PR #201, opened 2021-04-18
Added !group and !subdomain commands

snarain01:
Was talked about in rebuild a few days ago. !group [user] will show subdomain names tield to the user. !subdomain [subdomain] will show users tied to the subdomain.

Note: Had to rename plugins/github.py to plugins/github_info.py due to naming conflict


ja5087 (approved):
does what it says on the box

PR #202, opened 2021-04-19
changed trigger command and added mail and web hosting info

saurabhnarain:
Decided to change the trigger from "!group" and "!subdomain" to "create: group" and "create: subdomain" for consistency. Also added the ability for it to distinguish mail vhost and web vhost.

### DIFF
--- a/ircbot/plugin/group.py
+++ b/ircbot/plugin/group.py
@@ -44,8 +44,8 @@ def group_info(user):
         if user == web_vhost[1]['username']:
             web_domains.extend([web_vhost[0]] + web_vhost[1]['aliases'])
             continue
-    for i in mail_vhosts:
-        mail_domains.extend([i[1]])
+    for mail_vhost in mail_vhosts:
+        mail_domains.extend([mail_vhost[1]])
     if web_domains == []:
         web_domains = ['none']
     if mail_vhosts == set():
--- a/ircbot/plugin/group.py
+++ b/ircbot/plugin/group.py
@@ -7,8 +7,8 @@ from ocflib.vhost.web import get_vhosts
 
 
 def register(bot):
-    bot.listen(r'group (.*)$', group_lookup)
-    bot.listen(r'domain (.*)$', domain_lookup)
+    bot.listen(r'group (.*)$', group_lookup, require_mention=True)
+    bot.listen(r'domain (.*)$', domain_lookup, require_mention=True)
 
 
 def group_lookup(bot, msg):
--- a/ircbot/plugin/group.py
+++ b/ircbot/plugin/group.py
@@ -1,30 +1,67 @@
 """Print group account usernames from subdomain names and vice versa."""
+import re
+
+from ocflib.vhost.mail import get_mail_vhosts
+from ocflib.vhost.mail import vhosts_for_user
 from ocflib.vhost.web import get_vhosts
 
 
 def register(bot):
-    bot.listen(r'^!group(?: (.*))?', group_lookup)
-    bot.listen(r'^!subdomain(?: (.*))?', subdomain_lookup)
+    bot.listen(r'group (.*)$', group_lookup)
+    bot.listen(r'subdomain (.*)$', subdomain_lookup)
 
 
 def group_lookup(bot, msg):
     user = msg.match.group(1)
     vhosts = get_vhosts()
     subdomain = None
+    mail_vhost = vhosts_for_user(user) != set()
+    response = ''
     for vhost in vhosts.items():
         if user == vhost[1]['username']:
             subdomain = vhost[0]
-    if subdomain is None:
-        msg.respond('A subdomain does not exist with this group account.')
+            continue
+    if subdomain is None and not mail_vhost:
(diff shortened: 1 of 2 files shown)


ircbot/plugin/group.py
```diff
@@ -1,30 +1,67 @@
...
+        response = 'Web'
+    else:
+        user = mail_vhost_user
+        response = 'Mail'
+    msg.respond(response + ' virtual hosting at ' + subdomain + ' is enabled for account "' + user + '"', ping=False)
+
+
+def auto_complete(s):
+    valid_endings = ['.berkeley.edu', '.org', '.com', '.edu']
+    if not any([re.search(valid_ending, s) is not None for valid_ending in valid_endings]):
```
ja5087, line 65: I would probably write this as `all([search is None])` (this is DeMorgan's law), but it's up to you

ja5087: I recommend making the regex more strict or use `rfind` instead after `strip`ping the input, since you seem to be trying to search for a suffix. The `.` will match everything under the sun:

```
>>> re.search('.berkeley.edu', 'lberkeleyledu')
<re.Match object; span=(0, 13), match='lberkeleyledu'>
```

ircbot/plugin/group.py
```diff
@@ -1,30 +1,67 @@
...
+    subdomain = auto_complete(msg.match.group(1))
     user = get_vhosts().get(subdomain)
-    if user is None:
-        msg.respond('A group account does not exist with this subdomain.')
+    mail_vhost_user = None
+    response = ''
+    for mail_vhost in get_mail_vhosts():
+        if mail_vhost[1] == subdomain:
+            mail_vhost_user = mail_vhost[0]
+            continue
```
ja5087, line 47: same problem here, although most users do not have multiple mail vhosts

ja5087 (changes requested):
First pass on this. Thinking ahead, a lot of this functionality can be moved to ocflib at some point but that's not a blocker. Thanks for the changes! Looking forward to this in IRC.

ircbot/plugin/group.py
```diff
@@ -1,30 +1,67 @@
...
-    bot.listen(r'^!subdomain(?: (.*))?', subdomain_lookup)
+    bot.listen(r'group (.*)$', group_lookup)
+    bot.listen(r'subdomain (.*)$', subdomain_lookup)
 
 
 def group_lookup(bot, msg):
     user = msg.match.group(1)
     vhosts = get_vhosts()
     subdomain = None
+    mail_vhost = vhosts_for_user(user) != set()
```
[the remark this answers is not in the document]
snarain01: Yeah, I got rid of this part entirely

ircbot/plugin/group.py
```diff
@@ -1,30 +1,67 @@
...
+    response = ''
     for vhost in vhosts.items():
         if user == vhost[1]['username']:
             subdomain = vhost[0]
```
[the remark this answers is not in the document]
snarain01: I did not know that, the new logic also adds the aliases to the response

ircbot/plugin/group.py
snarain01: Yeah, rfind is definitely better! I didn't know it existed

ircbot/plugin/group.py
snarain01: Yeah, I found this interesting since the vhost-mail.conf has single line entries with only one domain and account. After some searching around I found a few accounts that had multiple vhosts (like uav). I was able to fix this too

snarain01:
Hey jaw! Thanks for looking through this. I pushed a new version that closely follows the format you recommended. I tested it out with multiple accounts and it seems to function well.

snarain01:
Also, I decided to change the "create: subdomain" to "create: domain" since we offer vhosting for non .berkeley.edu domains as well.

ircbot/plugin/group.py
```diff
@@ -1,30 +1,69 @@
...
+        mail_domains = ['none']
+    user_info_str = 'account: ' + user
+    is_group_account_str = ' | is group account: ' + str.lower(str(user_is_group(user)))
+    web_domains_str = ' | web vhosts: ' + ', '.join(web_domains)
+    mail_domains_str = ' | mail vhosts: ' + ', '.join(mail_domains)
+    return user_info_str + is_group_account_str + web_domains_str + mail_domains_str
+
+
+def auto_complete(s):
+    if get_tld('http://www.' + s, fail_silently=True) is None:
```
cg505, line 67: why `www.`?

saurabhnarain: Ah yeah, I thought that get_tld would only work if `www.` was there, but I just tested get_tld, and it works without the `www.`.

cg505 (approved):
I think you should try to test this on k8s if you haven't - I think you'll be fine but I'm not sure if some of these ocflib function will work correctly from within docker (e.g. if they expect certain local config files or something)

ja5087:
I already mounted the OCF local folder into the Docker container so it will probably work

ja5087 (approved):

PR #226, opened 2021-10-03
Add RT Summary Feature to IRCBot

saurabhnarain:
Worked on with lmathias during hack day. Definitely want to figure out how to efficiently get the rt number for the latest ticket so we don't waste time finding it. I wanted to merge this pr to do a preliminary test in the #service-comm channel.

### DIFF
--- a/ircbot/plugin/rt_summary.py
+++ b/ircbot/plugin/rt_summary.py
@@ -1,0 +1,42 @@
+"""Announce new RT list every day"""
+from ocflib.infra.rt import rt_connection
+from ocflib.infra.rt import RtTicket
+
+NUM_ITER = 10
+NUM_LIST = 10500
+CHANNEL = '#service-comm'
+
+
+def show_tickets(bot, date):
+    """Show RT tickets that need responses."""
+    rt = rt_connection(user='create', password=bot.rt_password)
+    bot.say(CHANNEL, f'Top 10 new tickets in the help queue for today ({str(date)}):')
+
+    counter = 0
+    counter_success = 0
+    num_newest_rt = newest_rt_number(rt, NUM_LIST)
+    while counter_success < NUM_ITER:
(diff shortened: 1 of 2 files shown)


ircbot/plugin/rt_summary.py
```diff
@@ -0,0 +1,42 @@
...
+from ocflib.infra.rt import rt_connection
+from ocflib.infra.rt import RtTicket
+
+NUM_ITER = 10
+NUM_LIST = 10500
+CHANNEL = '#service-comm'
+
+
+def show_tickets(bot, date):
+    """Show RT tickets that need responses."""
```
ethanhs, line 11: I thought about this a bit, and I think the heuristic for "needs responses most" are the *oldest* tickets in a queue.

If someone submits a ticket today, its usually ok if we take a day to respond, but if its been a week we really should respond ASAP.

ircbot/plugin/rt_summary.py
```diff
@@ -0,0 +1,42 @@
+"""Announce new RT list every day"""
+from ocflib.infra.rt import rt_connection
+from ocflib.infra.rt import RtTicket
+
+NUM_ITER = 10
+NUM_LIST = 10500
```
ethanhs, line 6: I think I see why this is 10500, but it probably deserves a comment above it on why it is 10500 (that's a recentish starting point to start searching from)

ircbot/plugin/rt_summary.py
saurabhnarain: that is a much better idea!

ircbot/plugin/rt_summary.py
saurabhnarain: yeah, hopefully we can figure out how to use the REST feature to find the most recent ticket instead. for now this should work!

PR #234, opened 2022-02-02
add debian lts security advisories

singingtelegram:
Since we are running the LTS version of Debian, we should follow LTS security advisories as well.

### DIFF
--- a/ircbot/plugin/debian_lts_security.py
+++ b/ircbot/plugin/debian_lts_security.py
@@ -1,0 +1,82 @@
+import operator
+import re
+from collections import namedtuple
+from datetime import datetime
+from typing import List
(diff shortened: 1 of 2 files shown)


singingtelegram:
semi-tested on dev

dkess:
This kind of code duplication (copying `debian_security.py`) is dangerous. If someone changes one file, they aren't going to remember to change the second one. And it's unlikely that a code reviewer would notice

There are a lot of ways you could pull this off without code duplication. One simple way: modify `debian_security.py` to look at a list of RSS feeds and loop through them.

singingtelegram:
Yeah, thanks for pointing that out!

